PR #2, opened 2020-03-11
Attachments2.0

Tigul:
# Added 
* New attachments that work when the robot is teleported
* Loose attachments that only work in one direction
* Refactor and clean up of the methods responsible for creating and managing the attachments

# Documentation 
If an attachment is created the transformation between the given link and the other object will be saved along with the link name and the flag if this is a loose attachment. In addition a PyBullet constraint will be created to ensure the constraint works during the simulation. 
The attachments will be resolved if either the `set_position_and_orientation` or `set_joint_state` of the object will be called.
If the attachment is loose only the object will be moved and none of the attached objects. Furthermore the attachment will be updated to take the new transformation into account.
For more details on how the methods work please consider the documentation in the methods itself.

**Please merge this pull request after #1 is merged**


tlipps:
Tested the attachments with the tests mentioned in this trello card https://trello.com/c/4SygwMWn/257-pycram-attachments-review .

PR #5, opened 2020-04-14
Multiverse

Tigul:
# Added
* Complete support for multiple BulletWorld
* A function to copy an existing BulletWorld

## Implementation
The support for more than one BulletWorld is largely done by adding the `physicClientID` argument to all statements which call PyBullet directly. Furthermore a function was added which creates a new BulletWorld and copies the state of the existing BulletWorld to the newly created one. This BulletWorld, however is in "direct mode" which means there is no graphical representation of the simulation. 

## Disclaimer
Because this PR is based on the work of the Pull Requests #2  #3 and #4 these PRs should be merged before this one.

### DIFF
--- a/pycram/src/pycram/bullet_world_reasoning.py
+++ b/pycram/src/pycram/bullet_world_reasoning.py
@@ -76,8 +76,6 @@ def stable(object, world=None):
     p.restoreState(state, physicsClientId=world_id)
     coords_prev = list(map(lambda n: round(n, 3), coords_prev))
     coords_past = list(map(lambda n: round(n, 3), coords_past))
-    print(coords_prev)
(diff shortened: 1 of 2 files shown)


pycram/src/pycram/bullet_world_reasoning.py
```diff
@@ -75,13 +76,15 @@ def stable(object, world=None):
     p.restoreState(state, physicsClientId=world_id)
     coords_prev = list(map(lambda n: round(n, 3), coords_prev))
     coords_past = list(map(lambda n: round(n, 3), coords_past))
+    print(coords_prev)
```
gaya-, line 79: prints

PR #12, opened 2020-09-09
Submodules

Tigul:
# Changed
Removed the macropy code from the src directory and moved it into an own repo, which is will be included as a submodule.


gaya-:
You need to apply the one commit that was required for PyCRAM:
https://github.com/dustin293/macropy
The macropy version and the version that we use are different.
Also, if you're getting the newest version of macropy you'd have to test if the CRAM Plan Language still works fine.

Tigul:
I know, i already did this. Concerning the newest version of macropy and the CRAM Plan Language, the last commit which changed the code of macropy was in may 2018 so this shouldn't be a problem.

PR #40, opened 2022-11-09
[pycram_demos/process_modules] deleted pepper_shopping demo and the real-world pr2_demo, and cleaned-up process modules

sunava:
[[pycram_demos] deleted pepper_shopping demo and the real-world pr2_demo](https://github.com/cram2/pycram/pull/40/commits/7b80f0e34f199a214e8c6fedc193d34f732e725b)


Tigul:
Could you move the process_module.py file back to src/pycram so the parent class is not in the same directory as the robot specific code.

PR #45, opened 2022-11-23
[designators] restructured designators

sunava:
The question here is; is there any fabulous import? idk, my idea would be to make everything from the designators available at all times since this is the main structure the demos are working with. No need to import them one by one? For now, I did it like this but this def breaks the task_tree demo I'm just checking the other demo all the time to see if it still runs. But within the task_tree there are still motion_designators and stuff, what do u think @Tigul?

### DIFF
--- a/demos/pycram_bullet_world_demo/demo.py
+++ b/demos/pycram_bullet_world_demo/demo.py
@@ -1,8 +1,9 @@
 import rospy
+import pycram.bullet_world_reasoning as btr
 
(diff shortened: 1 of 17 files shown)


sunava:
Meeting 24,Nov.22 -> @Tigul we decided to let designators like that so you can do from pycram.designators.AVIABLEDESIGANTORS imports which is nice. And task_tree demo stays here for now but we are to lazy to fix this demo.

PR #46, opened 2022-11-24
[extrernal_interfaces] added external_interface subfolder and merged all knowrob files into one file

sunava:


src/pycram/costmaps.py
```diff
@@ -5,7 +5,7 @@
 from pycram.robot_description import InitializedRobotDescription as robot_description
 from .bullet_world import BulletWorld
 from .bullet_world_reasoning import _get_images_for_target
-from .ik import request_ik, IKError
+from .external_interfaces.ik import request_ik, IKError
```
Tigul, line 8: Is this still used ? If not it should be removed

Tigul: Will be done in a later PR

sunava: no but this will be in one PR with all the other imports

PR #47, opened 2022-11-24
[robot_description] disassembled the specific robot description from the file and put them into new sub-categories. Added a robot_description_handler for the init and update functions. otherwise circular deps happens

sunava:


src/pycram/__init__.py
```diff
@@ -20,7 +20,7 @@
 import logging
 import logging.config
 import pycram.process_modules.available_process_modules
-
+import pycram.robot_descriptions.robot_description_handler
```
Tigul, line 23: This can be removed since the robot description needs to be explicitly imported if you want to use it.

PR #59, opened 2023-04-10
ORM package for pycram.

tomsch420:
ORM package for pycram including tests, examples and documentation.


src/pycram/designators/action_designator.py
```diff
@@ -16,10 +16,14 @@
            "CloseAction"]
 
 from typing import List
+import sqlalchemy.orm
+import pycram.orm.action_designator
```
Tigul, line 20: Relative import

PR #70, opened 2023-04-22
Suppressed warnings on import.

tomsch420:
It originates from line 544 in urdf_parser_py/xml_reflection/core.py

It seems to be the case, that urdf_parser does not know what a sensor is, hence it raises the error. This cannot be fixed from pycram, therefore it gets supressed in the init.py. (see https://github.com/ros/urdf_parser_py/issues/44)


Tigul:
Please clean up the pull request, there are a lot of edited files that have nothing to do with the purpose of the PR. Also why did you create a new utils file and did not use the helper file ?

PR #89, opened 2023-08-11
Add tiago_dual urdf & description. Cleanup launch.

artnie:
Launcher
* move ik nodes to pr2 group
* add tiago
* add robot state publisher arg

URDF
* add compiled tiago urdf, like loaded here: 
https://github.com/cram2/cram/blob/bb7df4fbe50099693eec27fa290d6e682effdde4/cram_demos/cram_projection_demos/launch/everything.launch#L50

Description
* add tiago description based on this: 
https://github.com/cram2/cram/blob/devel/cram_demos/cram_robot_descriptions/cram_tiago_description/src/knowledge.lisp


artnie:
How to test

change ik_and_description.launch to load tiago
```
  <arg name="robot" default="tiago_dual" />
  <arg name="use_knowrob" default="false"/>
  <arg name="publish_robot" default="True"/>
```
Launch `roslaunch pycram ik_and_description.launch`

Open the examples/action_designator.ipynb in pycharm. 
Change the first code cell to
```python
from pycram.bullet_world import BulletWorld, Object
from pycram.pose import Pose

world = BulletWorld()
tiago = Object("tiago", "tiago_dual", "tiago_dual.urdf")
```

artnie:
Btw, the robot is called `tiago_dual` because that's the name in the URDF.

Tigul:
Is the robot_state_publisher in the launch file necessary ? Since there is a class in PyCRAM which also publishes the current JointState of the robot. This might cause problems

artnie:
> Is the robot_state_publisher in the launch file necessary ? Since there is a class in PyCRAM which also publishes the current JointState of the robot. This might cause problems

I had the issue, that the robots TF wasn't published properly, so I added the robot state publisher. Publishing TF will be required for working online with Rvizweb

artnie:
And it's optional, I put the arg to false by default

artnie:
Should I split the joint state publisher and robot state publisher into 2 separate groups in the launch file and habe an arg for each?

Tigul:
The joint_state_publisher and robot_state_publisher just publish the static information from the URDF, so it might be nicer to expose the joint_state_broadcaster and TF broadcaster of PyCRAM as ROS nodes and use them instead

artnie:
> The joint_state_publisher and robot_state_publisher just publish the static information from the URDF, so it might be nicer to expose the joint_state_broadcaster and TF broadcaster of PyCRAM as ROS nodes and use them instead

Removed the joint- and robot_state_publisher for now.

PR #94, opened 2023-08-28
Make pycram dockerizable

HoangGiang93:
I fixed the apartment and tiago_dual so that only collision meshes are considered and can be opened locally.
`os.getlogin()` is replaced with `getpass().getuser()`, see https://stackoverflow.com/questions/4399617/invalid-argument-when-running-os-getlogin-from-cron


Tigul:
A few remarks: 
  * I don't think we should put the mesh files in the resource folder, like with the other robots there should be a dependecy for that. They will propably not used because these are not ROS packages
      * You can add the iai_tiago (https://github.com/code-iai/iai_tiago/) to the ROS install file in the root directory of the repo. 
   * This is more a subjective opinion but I don't like the PR2 without the visual tags, since it does not seem to load significantly faster.

Tigul:
I'm closing this since there happened nothing for the past months.

PR #97, opened 2023-09-08
[]examples] added a very small jypter nbook

sunava:
this is for the  local_transformer of pycram


sunava:
@Tigul I will accept this for now since it is only a jypter notebook and our event is coming up

PR #103, opened 2023-10-24
[action-desig-pickup]added everything for the pickup similar to cram

sunava:
the special knowledge needs a better place and the pr_1 link is a bit annoying maybe we could find a better way for that.

### DIFF
--- a/src/pycram/designator.py
+++ b/src/pycram/designator.py
@@ -572,7 +572,16 @@ class LocationDesignatorDescription(DesignatorDescription):
         """
         raise NotImplementedError(f"{type(self)}.ground() is not implemented.")
 
-
+#this knowledge should be somewhere else i guess
+SPECIAL_KNOWLEDGE = {
+    'bigknife':
+        [("top", [-0.08, 0, 0])],
+    'whisk':
+        [("top", [-0.08, 0, 0])],
+    'bowl':
+        [("front", [1.0, 2.0, 3.0]),
+         ("key2", [4.0, 5.0, 6.0])]
+}
 class ObjectDesignatorDescription(DesignatorDescription):
     """
     Class for object designator descriptions.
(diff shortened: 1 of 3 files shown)


src/pycram/designator.py
```diff
@@ -679,6 +688,44 @@ def __repr__(self):
                 [f"{f.name}={self.__getattribute__(f.name)}" for f in dataclasses.fields(self)] + [
                     f"pose={self.pose}"]) + ')'
 
+        def special_knowledge_adjustment_pose(self, grasp, pose):
+            """
+            Returns the adjusted target pose based on special knowledge for "grasp front".
+            """
+
```
Tigul, line 695: You could also to the transform to object frame here to ensure that the pose is in the correct frame.

src/pycram/designators/action_designator.py
```diff
@@ -294,8 +294,72 @@ class Action(ActionDesignatorDescription.Action):
...
+                                       grasp)
+
+            # Set the orientation of the object pose by grasp in MAP
+            adjusted_oTm.orientation.x = ori[0]
+            adjusted_oTm.orientation.y = ori[1]
+            adjusted_oTm.orientation.z = ori[2]
+            adjusted_oTm.orientation.w = ori[3]
+
+            # prepose depending on the gripper (its annoying we have to put pr2_1 here tbh
+            gripper_frame = "pr2_1/l_gripper_tool_frame" if self.arm == "left" else "pr2_1/r_gripper_tool_frame"
```
Tigul, line 325: This would only work in very specific cases. Please use: robot.get_link_tf_frame(robot_description.get_tool_frame(self.arm))

src/pycram/designators/action_designator.py
```diff
@@ -294,8 +294,72 @@ class Action(ActionDesignatorDescription.Action):
...
+            print(adjusted_oTm)
```
Tigul, line 326: should this print be here ?

src/pycram/designators/action_designator.py
```diff
@@ -294,8 +294,72 @@ class Action(ActionDesignatorDescription.Action):
...
+            # First rotate the gripper, so the further calculations maakes sense
+            tmp_for_rotate_pose = object.local_transformer.transform_pose(adjusted_oTm, gripper_frame)
+            print(tmp_for_rotate_pose)
```
Tigul, line 329: another print

src/pycram/designators/action_designator.py
```diff
@@ -314,9 +378,7 @@ def insert(self, session: sqlalchemy.orm.session.Session, **kwargs):
 
             return action
 
-    def __init__(self,
-                 object_designator_description: Union[ObjectDesignatorDescription, ObjectDesignatorDescription.Object],
-                 arms: List[str],
+    def __init__(self, object_designator_description: ObjectDesignatorDescription, arms: List[str],
```
Tigul, line 381: Why did you change the typing?

src/pycram/designators/action_designator.py
```diff
@@ -339,11 +400,7 @@ def ground(self) -> Action:
 
         :return: A performable designator
         """
-        obj_desig = self.object_designator_description if isinstance(self.object_designator_description,
-                                                                     ObjectDesignatorDescription.Object) else self.object_designator_description.resolve()
-
-        return self.Action(obj_desig, self.arms[0], self.grasps[0])
-
+        return self.Action(self.object_designator_description.ground(), self.arms[0], self.grasps[0])
```
Tigul, line 403: Why not allow object designator descriptions ?

src/pycram/helper.py
```diff
@@ -169,3 +169,59 @@ def has(self, index: int) -> bool:
             return True
         except StopIteration:
             return False
+def axis_angle_to_quaternion(axis, angle):
```
Tigul, line 172: Please add typing

src/pycram/helper.py
```diff
@@ -169,3 +169,59 @@ def has(self, index: int) -> bool:
...
+    """
+    Convert axis-angle to quaternion.
+    axis: (x, y, z) tuple representing rotation axis.
+    angle: rotation angle in degree
+    """
```
Tigul, line 177: Please look at the other functions for the doc string. There needs to be a new line between the description and parameter description. Parameter need to be written as :param 'name': ... and return as :return: 
Otherwise the auto generation for the API would not work

Tigul (changes requested):
Please change

src/pycram/designator.py
```diff
@@ -572,7 +572,16 @@ def ground(self) -> Location:
         """
         raise NotImplementedError(f"{type(self)}.ground() is not implemented.")
 
-
+#this knowledge should be somewhere else i guess
```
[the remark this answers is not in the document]
sunava: as we discussed in person i will let this here

src/pycram/designator.py
```diff
@@ -572,7 +572,16 @@ def ground(self) -> Location:
...
+SPECIAL_KNOWLEDGE = {
+    'bigknife':
+        [("top", [-0.08, 0, 0])],
```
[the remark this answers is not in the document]
sunava: @Tigul id say we just let it like this since ull change the interface anyway?

PR #112, opened 2023-11-21
[failure_handling] rewrote failure handling, added test cases

sunava:


src/pycram/failure_handling.py
```diff
@@ -0,0 +1,93 @@
...
+        perform(): Abstract method
+    """
+
+    def __init__(self, designator_description:DesignatorDescription):
+        """
+        Initializes a new instance of the FailureHandling class.
+
+        Args:
+            designator_description (DesignatorDescription): The description or context
+            of the task or process for which the failure handling is being set up.
```
Tigul, line 25: Could you do the dsignator_description in one line otherwise the string is not rendered correctly

PR #114, opened 2023-11-28
[Giskard Interface Notebook] created joint publisher from giskard res…

AbdelrhmanBassiouny:
…ult to sync joints, with minor changes in robot_state_updater.py


Tigul:
With SemRoCo/giskardpy#129 giskard has now a built-in joint state publisher. So I would close this PR and update the Giskard example with the robot_state_updater

PR #117, opened 2023-12-13
Added/updated querying guides and fixed orm usage in different places

davidprueser:


tomsch420 (commented):
Everything besides the one unnecessary cell if great!

tomsch420 (approved):

PR #125, opened 2024-01-11
Bugfix retract pose

Leusmann:
Fixes a bug where the retract_pose was calculated within the wrong frame (world). Further more unifies the retract offset to 0.07 (7cm)

Additionally resets the joint state of the robot after each pose generation

### DIFF
--- a/src/pycram/pose_generator_and_validator.py
+++ b/src/pycram/pose_generator_and_validator.py
@@ -151,13 +151,12 @@ def reachability_validator(pose: Pose,
     if robot in allowed_collision.keys():
         allowed_robot_links = allowed_collision[robot]
 
+    joint_state_before_ik=robot._current_joint_states
     try:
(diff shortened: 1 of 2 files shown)


src/pycram/pose_generator_and_validator.py
```diff
@@ -155,7 +155,9 @@ def reachability_validator(pose: Pose,
         # resp = request_ik(base_link, end_effector, target_diff, robot, left_joints)
         resp = request_ik(target, robot, left_joints, left_gripper)
 
+        joint_state_before_ik=robot._current_joint_states
         _apply_ik(robot, resp, left_joints)
+        robot.set_joint_states(joint_state_before_ik)
```
Tigul, line 160: You are resetting the robot joint state before the collision check is done. Please move the code in line 160 after the collision was checked.

PR #129, opened 2024-01-18
Updates reachabiltiy_validator to check for all tool_frames

Leusmann:
Updates the reachability_validator to check all tool_frames for reach ability, keeps the retrace pose in mind


Leusmann:
Nice find. I can reproduce the error on my side. Will fix

Leusmann:
I think I found a solution for this problem, which is a bit surprising. The IK Server only finds solutions if the position was given in the world frame? 
Could somebody please check if it works for them as well?

src/pycram/pose_generator_and_validator.py
```diff
@@ -153,36 +154,59 @@ def reachability_validator(pose: Pose,
...
         except IKError:
             pass
         finally:
             robot.set_joint_states(joint_state_before_ik)
+    if arms:
+        res = True
     return res, arms
+
+
+def collision_check(robot, allowed_collision):
```
Tigul, line 192: Doc please.

Leusmann:
Added some documentation strings - should be better now

PR #134, opened 2024-02-23
Added auto inserts for action designator orms and fixed querying from occupancy costmap

davidprueser:


src/pycram/designators/actions/actions.py
```diff
@@ -32,14 +40,24 @@ def perform(self) -> None:
         """
         pass
 
-    @abc.abstractmethod
     def to_sql(self) -> Action:
         """
```
tomsch420, line 44: add some description in the doc string on what requirements are needed to be able to do this automatically. And also when and how to overwrite it

tomsch420 (changes requested):
Besides the docstring issues everything is great.

tomsch420 (approved):

PR #137, opened 2024-03-14
Updated the CI pipeline

leonyi4:
Please make a new docker repository and add the docker image tag as DOCKER_IMAGE under GitHub variables.


.github/workflows/new-pycram-ci.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Install pytest, pyjpt & mlflow
+        run: |
+          pip3 install --ignore-installed pytest pyjpt mlflow
+
+      - name: Run tests
+        run: |
+          sudo -s
+          source /opt/ros/noetic/setup.bash
+          cd /opt/ros/overlay_ws
```
Tigul, line 72: What do the first three statements, wouldn't the source, roscd and pytest be enough?

.github/workflows/update-docker-image.yml
```diff
@@ -0,0 +1,37 @@
+name: update_docker_image
+on:
+  workflow_run:
+    workflows: ["pycram_docker_ci"]
+    types:
+      - completed
+  workflow_dispatch: # manual run
+    inputs:
```
Tigul, line 8: The Dockerfile only needs to be updated if the requirements change. We could also automate this: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files

Tigul (changes requested):

PR #140, opened 2024-04-04
Stretch

Tigul:
Updates to the stretch integration, this includes the correct URDF and corresponding robot description, process modules for the real Stretch and a method to force an orientation generator for PoseGenerators.


src/pycram/pose_generator_and_validator.py
```diff
@@ -17,6 +17,13 @@
 
 class PoseGenerator:
```
tomsch420, line 18: class documentation for what this class is supposed to do is missing.

src/pycram/pose_generator_and_validator.py
```diff
@@ -17,6 +17,13 @@
...
     current_orientation_generator = None
+    """
+    If no orientation generator is given, this generator is used to generate the orientation of the robot.
+    """
+    override_orientation_generator = None
```
tomsch420, line 23: why are these 2 different fields? Just default it to the ordinary orientation generate and if the user wants a different one he can set that, or am I missing something?

src/pycram/process_modules/stretch_process_modules.py
```diff
@@ -29,37 +39,84 @@ def _execute(self, designator) -> Any:
...
+        desig.object_part.bullet_world_object.set_joint_state(container_joint,
+                                                              part_of_object.get_joint_limits(
+                                                                  container_joint)[1] - 0.05)
 
 
 class StretchClose(ProcessModule):
-    def _execute(self, designator) -> Any:
-        pass
+    """
+    Process module for the simulated Stretch that closes an already grasped container
```
tomsch420, line 104: duplicated string

tomsch420 (changes requested):

tomsch420:
And where tests?

tomsch420 (approved):

PR #148, opened 2024-04-22
[resources]: Added additional resources

K3cks:
Added additional models for objects, robots and worlds. The objects are a smaller stl version of their 3d models from https://github.com/Multiverse-Framework/Multiverse-Assets/tree/main/unchecked/objects .
Restructured objects, robots and worlds into seperate directories. 
Adjusted default paths for robots and search for resources in all subdirectories.


sunava:
Please fix tests aswell.

resources/models/defenitly_not_hard_coded.pm
tomsch420, line 1: Please do not commit models to git. This model is small and hence wont blow up, however it is not the intended functionality. Use some MLOps stack like MLFLow that is set up in the institute for specifically that purpose.

tomsch420 (changes requested):
I am very sure that all resource files should not be tracked in git. Talk to @Tigul  about his vision for that.

resources/models/defenitly_not_hard_coded.pm
sunava: Mhm, I was thinking that in the future, we could have models like yours to answer questions like, "where might I find the spoon?" In my demos (which haven't been pushed yet because I'm focused on finalizing the functionality), I'm already using that model. I thought this could be the first step toward using your model, because the calls remain the same; only your answers would change as your model gets more trained.

sunava: oh i see, that was the wrong model nvm thanks for noticing

sunava (approved):

sunava:
waiting for the meeting tmr to approve this. but looks good

Tigul:
Maybe we could make a seperate Repo and use it as a submodule for the meshes.

sunava:
will rework this based on the discussion today

PR #150, opened 2024-04-22
[hsrb] changed naming and some description parts

sunava:
Since our HSR is also named "HSRB" on the real parameter server, I updated all references from "HSR" to "HSRB." Additionally, I corrected some sections with inaccurate descriptions.


src/pycram/robot_descriptions/hsrb_description.py
```diff
@@ -25,25 +24,29 @@ def __init__(self):
         # Neck
         neck_links = ["head_pan_link", "head_tilt_link"]
         neck_joints = ["head_pan_joint", "head_tilt_joint"]
-        neck_forward = {"forward": [0.0, 0.0], "down": [0.0, -0.7]}
+        neck_forward = {"forward": [0.0, 0.0], "down": [0.0, 0, -0.7]}
```
Tigul, line 27: This might cause problems since the neck chain only has two joints, but I don't think the neck_forward or neck_down poses are used anywhere so this should still be fine.

PR #152, opened 2024-04-22
[world] Added a function to retrieve all objects except for robots 

sunava:
and environmental elements, along with a function that spawns a rigid body. Using the rigid body function to spawn objects based on the perceived unknown objects from rk.


Tigul:
I would like to handle the spawning of Viisual shapes differently as discussed this morning.

sunava:
> I would like to handle the spawning of Viisual shapes differently as discussed this morning.

will change the pr as discussed

PR #161, opened 2024-05-07
Full body giskard

Tigul:
Adds giskard as full body motion solver. This PR still needs work on the giskard interface for compliance with parallel execution


sunava (approved):
Looks good and works on my computer we just need to work on the parallel as u already described in the description

Tigul:
Parallel execution is fixed, I just forget to adjust the comment

PR #172, opened 2024-07-19
Optimization 24

Tigul:
Some small optimization for Pose and Transformations. 
* The quaternion normalization of Pose was changed to list comprehention
* When transforming a Pose only the affected objects will update their transforms.

### DIFF
--- a/src/pycram/datastructures/world.py
+++ b/src/pycram/datastructures/world.py
@@ -263,12 +263,12 @@ class World(StateEntity, ABC):
         """
         pass
 
-    def get_object_by_name(self, name: str) -> List[Object]:
+    def get_object_by_name(self, name: str) -> Optional[Object]:
         """
-        Returns a list of all Objects in this World with the same name as the given one.
+        Returns the object with the given name. If there is no object with the given name, None is returned.
 
         :param name: The name of the returned Objects.
-        :return: A list of all Objects with the name 'name'.
+        :return: The object with the given name, if there is one.
         """
 
         object = list(filter(lambda obj: obj.name == name, self.objects))
--- a/src/pycram/datastructures/world.py
+++ b/src/pycram/datastructures/world.py
@@ -270,7 +270,11 @@ class World(StateEntity, ABC):
         :param name: The name of the returned Objects.
         :return: A list of all Objects with the name 'name'.
         """
-        return list(filter(lambda obj: obj.name == name, self.objects))[0]
+
+        object = list(filter(lambda obj: obj.name == name, self.objects))
+        if len(object) > 0:
+            return object[0]
+        return None
 
     def get_object_by_type(self, obj_type: ObjectType) -> List[Object]:
         """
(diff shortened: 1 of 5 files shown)


src/pycram/datastructures/world.py
```diff
@@ -270,7 +270,11 @@ def get_object_by_name(self, name: str) -> List[Object]:
         :param name: The name of the returned Objects.
         :return: A list of all Objects with the name 'name'.
         """
-        return list(filter(lambda obj: obj.name == name, self.objects))[0]
+
+        object = list(filter(lambda obj: obj.name == name, self.objects))
```
tomsch420, line 274: The docstring here looks off.

src/pycram/local_transformer.py
```diff
@@ -64,6 +64,19 @@ def transform_to_object_frame(self, pose: Pose,
             target_frame = world_object.tf_frame
         return self.transform_pose(pose, target_frame)
 
+    def update_transforms_for_objects(self, source_object_name: str, target_object_name: str) -> None:
+        """
+        Updates the transforms for objects affected by the transformation. The objects are identified by their names.
```
tomsch420, line 69: add the word inplace to the docstring such that its clearer

src/pycram/local_transformer.py
```diff
@@ -64,6 +64,19 @@ def transform_to_object_frame(self, pose: Pose,
...
+        :param source_object_name: Name of the object of the source frame
+        :param target_object_name: Name of the object of the target frame
+        """
+        source_object = self.world.get_object_by_name(source_object_name)
+        target_object = self.world.get_object_by_name(target_object_name)
+        for obj in {source_object, target_object}:
+            if obj:
+                obj.update_link_transforms()
+
     def transform_pose(self, pose: Pose, target_frame: str) -> Union[Pose, None]:
```
tomsch420, line 80: Union[Pose, None] is better described by Optional[Pose]

src/pycram/local_transformer.py
```diff
@@ -93,9 +107,13 @@ def lookup_transform_from_source_to_target_frame(self, source_frame: str, target
         Update the transforms for all world objects then Look up for the latest known transform that transforms a point
          from source frame to target frame. If no time is given the last common time between the two frames is used.
 
+        :param source_frame: The frame in which the point is currently represented
+        :param target_frame: The frame in which the point should be represented
         :param time: Time at which the transform should be looked up
+        :return: The transform from source_frame to target_frame
         """
-        self.world.update_transforms_for_objects_in_current_world()
+        self.update_transforms_for_objects(source_frame.split("/")[0], target_frame.split("/")[0])
```
tomsch420, line 115: this name split should be a function. Consider that other people are doing something similar and are wondering how to extract the relevatn part. Perhaps they do it slightly wrong in their own way. A common function for it ensures long lasting support

tomsch420 (changes requested):

PR #174, opened 2024-07-25
Add dual arm pickup designator

J-Schaefer:
Added a specialized action designator that chooses the closest gripper to an object to grasp more efficiently.


src/pycram/designators/specialized_designators/action/dual_arm_pickup_action.py
```diff
@@ -0,0 +1,88 @@
...
+                         arms=[Arms.LEFT, Arms.RIGHT],
+                         grasps=grasps,
+                         resolver=resolver,
+                         ontology_concept_holders=ontology_concept_holders)
+
+        self.object_designator_description: Union[
+            ObjectDesignatorDescription, ObjectDesignatorDescription.Object] = object_designator_description
+
+        left_gripper_frame = World.robot.get_link_tf_frame(
+            RobotDescription.current_robot_description.kinematic_chains.get('left').get_tool_frame())
```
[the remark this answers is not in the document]
Tigul: You can also use "get_manipulator_chains" this will return all Kinematic chains which have an end effector

src/pycram/designators/specialized_designators/action/dual_arm_pickup_action.py
```diff
@@ -0,0 +1,88 @@
...
+                                                                                            object_frame)
+            object_V_gripper: Vector3 = object_T_gripper.translation  # translation vector
+            distance = norm(array([object_V_gripper.x, object_V_gripper.y, object_V_gripper.z]))
+            rospy.loginfo(f"Distance between {gripper} and {obj_desig.name}: {distance}")
+            distances.append(distance)
+
+        min_index = distances.index(min(distances))
+        winner_frame = self.gripper_list[min_index]
+
+        if winner_frame == World.robot.get_link_tf_frame(
```
Tigul, line 77: The kinematic chain description contains the arm_type (i.e. Arms.LEFT or Arms.RIGHT) it would be better to use this to avoid this if-else block. You would need to save the KinemacticChainDescriptions in the gripper_list, though

src/pycram/designators/specialized_designators/action/dual_arm_pickup_action.py
```diff
@@ -0,0 +1,88 @@
...
+            winner = Arms.LEFT
+            rospy.loginfo("Returning left arm as the winner")
+        elif winner_frame == World.robot.get_link_tf_frame(
+                RobotDescription.current_robot_description.kinematic_chains.get('right').get_tool_frame()):
+            winner = Arms.RIGHT
+            rospy.loginfo("Returning right arm as the winner")
+        else:
+            raise ValueError(f"Could not determine the winner arm for the pickup action. Winner: {winner_frame}")
+
+        return PickUpActionPerformable(object_designator=obj_desig, arm=winner, grasp=self.grasps[0])
```
Tigul, line 88: Just a general thing, you implement this in the ground method which is the default for reasoning. It would be better to move your code to the \_\_iter\_\_  method and yield the results according to the calculated distances.

Tigul (changes requested):

Tigul (approved):

PR #179, opened 2024-08-07
[failure handling]  A subclass of FailureHandling

sunava:
Introduces the RetryMonitor class, a subclass of FailureHandling, which implements a retry mechanism specifically designed to work with a Monitor. This class aims to improve the robustness of task execution by allowing retries in case the monitoring condition is triggered.

### DIFF
--- a/src/pycram/failure_handling.py
+++ b/src/pycram/failure_handling.py
@@ -1,7 +1,8 @@
+from .datastructures.enums import State
 from .designator import DesignatorDescription
 from .plan_failures import PlanFailure
 from threading import Lock
-from typing import Union
+from typing_extensions import Union, Tuple, Any, List
 from .language import Language, Monitor
 
 
@@ -18,7 +19,8 @@ class FailureHandling(Language):
         """
         Initializes a new instance of the FailureHandling class.
 
-        :param designator_description: The description or context of the task or process for which the failure handling is being set up.
+        :param Union[DesignatorDescription, Monitor] designator_description: The description or context of the task
+        or process for which the failure handling is being set up.
         """
         self.designator_description = designator_description
 
@@ -40,15 +42,10 @@ class Retry(FailureHandling):
 
     This class represents a specific failure handling strategy where the system
     attempts to retry a failed action a certain number of times before giving up.
-
-    Attributes:
-        max_tries (int): The maximum number of attempts to retry the action.
-
-    Inherits:
-        All attributes and methods from the FailureHandling class.
-
-    Overrides:
-        perform(): Implements the retry logic.
+    """
+    max_tries: int
+    """
+    The maximum number of attempts to retry the action.
     """
 
     def __init__(self, designator_description: DesignatorDescription, max_tries: int = 3):
@@ -61,7 +58,7 @@ class Retry(FailureHandling):
         super().__init__(designator_description)
         self.max_tries = max_tries
 
-    def perform(self):
+    def perform(self) -> Tuple[State, List[Any]]:
         """
         Implementation of the retry mechanism.
 
@@ -88,27 +85,22 @@ class RetryMonitor(FailureHandling):
 
     This class represents a specific failure handling strategy that allows us to retry a demo that is
     being monitored, in case that monitoring condition is triggered.
-
-    Attributes:
-        max_tries (int): The maximum number of attempts to retry the action.
-        recovery (dict): A dictionary that maps exception types to recovery actions
-
-    Inherits:
-        All attributes and methods from the FailureHandling class.
-
-    Overrides:
-        perform(): Implements the retry logic.
+    """
+    max_tries: int
+    """
+    The maximum number of attempts to retry the action.
+    """
+    recovery: dict
+    """
+    A dictionary that maps exception types to recovery actions
     """
 
     def __init__(self, designator_description: Monitor, max_tries: int = 3, recovery: dict = None):
         """
-        Initializes a new instance of the Retry class.
-
-        Args:
-            designator_description (DesignatorDescription): The description or context
-            of the task or process for which the retry mechanism is being set up.
-            max_tries (int, optional): The maximum number of attempts to retry. Defaults to 3.
-            recovery (dict, optional): A dictionary that maps exception types to recovery actions. Defaults to None.
+        Initializes a new instance of the RetryMonitor class.
+        :param Monitor designator_description: The Monitor instance to be used.
+        :param int max_tries: The maximum number of attempts to retry. Defaults to 3.
+        :param dict recovery: A dictionary that maps exception types to recovery actions. Defaults to None.
         """
         super().__init__(designator_description)
         self.max_tries = max_tries
@@ -126,37 +118,37 @@ class RetryMonitor(FailureHandling):
                     raise TypeError("Values in the recovery dictionary must be instances of the Language class.")
             self.recovery = recovery
 
-    def perform(self):
+    def perform(self) -> Tuple[State, List[Any]]:
         """
-        Implementation of the retry mechanism.
+        This method attempts to perform the Monitor + plan specified in the designator_description. If the action
+        fails, it is retried up to max_tries times. If all attempts fail, the last exception is raised. In every
+        loop, we need to clear the kill_event, and set all relevant 'interrupted' variables to False, to make sure
+        the Monitor and plan are executed properly again.
 
-        This method attempts to perform the Monitor + plan specified in the designator_description.
-        If the action fails, it is retried up to max_tries times. If all attempts fail,
-        the last exception is raised. In every loop, we need to clear the kill_event, and set all
-        relevant 'interrupted' variables to False, to make sure the Monitor and plan are executed
-        properly again
+        :raises PlanFailure: If all retry attempts fail.
 
-        Raises:
-            PlanFailure: If all retry attempts fail.
-
-        Returns:
-            The state of the execution performed, as well as a flattened list of the results, in the correct order
+        :return: The state of the execution performed, as well as a flattened list of the
+        results, in the correct order
         """
 
         def reset_interrupted(child):
-            if hasattr(child, "interrupted"):
-                child.interrupted = False
-            for sub_child in getattr(child, "children", []):
-                reset_interrupted(sub_child)
+            child.interrupted = False
+            try:
+                for sub_child in child.children:
+                    reset_interrupted(sub_child)
+            except AttributeError:
+                pass
 
         def flatten(result):
             flattened_list = []
-            for item in result:
-                if isinstance(item, list):
-                    flattened_list.extend(item)
-                else:
-                    flattened_list.append(item)
-            return flattened_list
+            if result:
+                for item in result:
+                    if isinstance(item, list):
+                        flattened_list.extend(item)
+                    else:
+                        flattened_list.append(item)
+                return flattened_list
+            return None
 
         status, res = None, None
         with self.lock:
--- a/src/pycram/failure_handling.py
+++ b/src/pycram/failure_handling.py
@@ -1,5 +1,8 @@
 from .designator import DesignatorDescription
 from .plan_failures import PlanFailure
+from threading import Lock
+from typing import Union
+from .language import Language, Monitor
 
 
 class FailureHandling:
@@ -11,7 +14,7 @@ class FailureHandling:
     to be extended by subclasses that implement specific failure handling behaviors.
     """
 
-    def __init__(self, designator_description: DesignatorDescription):
+    def __init__(self, designator_description: Union[DesignatorDescription, Monitor]):
         """
         Initializes a new instance of the FailureHandling class.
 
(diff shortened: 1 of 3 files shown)


src/pycram/failure_handling.py
```diff
@@ -79,5 +82,75 @@ def perform(self):
...
+                child.interrupted = False
+            for sub_child in getattr(child, "children", []):
+                reset_interrupted(sub_child)
+
+        def flatten(result):
+            flattened_list = []
+            for item in result:
+                if isinstance(item, list):
+                    flattened_list.extend(item)
+                elif isinstance(item, tuple):
```
artnie, line 135: reduce to one 'else' please. The second and 'otherwise' case are the same.

artnie (commented):
Please provide a minimal working example or at least a code snippet on how to extend an action perfom with this monitor. Then I can test and approve.

sunava:
Adds "RetryMonitor", which retries a plan up until max_tries times, while refreshing the monitoring of that plan. An example of that can be found in "examples/retry_monitor.ipynb", which showcases how RetryMonitor is to be used, and also how to manually update variables at runtime while using RetryMonitor.

```Python
# This line uses the already existing Monitor expression in PyCram
plan = ParkArmsAction([Arms.BOTH]) + MoveTorsoAction([0.25]) >> Monitor(monitor)

# Now we can use the RetryMonitor expression to retry our plan up to max_tries times.
state, [ParkArmsReturnValue, MoveTorsoReturnValue] = RetryMonitor(plan, max_tries=5).perform()
```

The .perform() method of the Monitor class was adjusted slightly to allow for smooth and consistent refreshing of the monitor, as the old "check_condition" implementation seemed to sometimes cause the condition not to be checked properly after it was refreshed, due to the sleep loop.

Furthermore, it is now Raising a PlanFailure error if the condition is met, which is caught by the RetryMonitor.

Lastly, the class FailureHandling now inherits from Language, to allow passing instances of FailureHandling to be monitored as well, and designator_description is now expected to be "Union[DesignatorDescription, Monitor]".

---

Description of the changes:
The state is the state of execution specified in the documentation, and List[Any] is a list of results of the children executed in the correct order. This allows us to use RetryMonitor as follows, given that "sleep" and "print_height" simply return their parameters after execution:

```Python
s_time = Code(lambda: sleep(sleep_time))
p_height = Code(lambda: print_height(height))
move_torso = Code(lambda: MoveTorsoAction([height]).resolve().perform())

plan = (p_height + move_torso + s_time) * 2 >> Monitor(monitor)
h, _, s = RetryMonitor(plan, max_tries=5).perform()
print(f"Used height {h}")
print(f"Used sleep time {s}")
```

While this should not change the behaviour or usage of these functions if you have not used their return values thus far, if you did, you will need to adjust them accordingly, to correctly unpack the return values. If before for example you used:

```Python
state = SomeExpression + SomeOtherExpression
```

You should now use:

```Python
state, _ = SomeExpression + SomeOtherExpression
```

---

We can define different recovery behaviours for different error messages as follows:
```Python
recover1 = Code(lambda: recovery1())
recover2 = Code(lambda: recovery2())

, [, milk_desig] = RetryMonitor(plan, max_tries=5, recovery={NotALanguageExpression: recover1,
PerceptionObjectNotFound: recover2}).perform()
```

src/pycram/language.py
```diff
@@ -260,34 +261,50 @@ def __init__(self, condition: Union[Callable, Fluent] = None):
...
         t = threading.Thread(target=check_condition)
         t.start()
-        res = self.children[0].perform()
-        self.kill_event.set()
-        t.join()
-        return res
+        try:
+            state, result = self.children[0].perform()
+            if not self.exception_queue.empty():
+                print("Raising PlanFailure")
```
artnie, line 302: When running the tests, this line is executed 12 times, while max_retries=6.
Is this intended? If so, please remove this print.

src/pycram/language.py
```diff
@@ -260,34 +261,50 @@ def __init__(self, condition: Union[Callable, Fluent] = None):
...
-                return
-            for child in self.children:
-                child.interrupt()
+            while not self.kill_event.is_set():
+                try:
+                    cond = self.condition.get_value()
+                    if cond:
+                        for child in self.children:
+                            if hasattr(child, 'interrupt'):
+                                child.interrupt()
```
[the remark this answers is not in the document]
LucaKro: yes it is intended, but its not a decision that i have made. RetryMonitor uses the already existing [Monitor](https://pycram.readthedocs.io/en/latest/notebooks/language.html#monitor) that was added by (i think) Jonas. here is the Doc for the monitor: 
"Monitor allows to monitor the execution of a language expression and interrupt it as soon as a given condition is fulfilled. The condition can either be a Callable which returns a boolean or a Fluent. When executed the Monitor will create a separate thread which will check if the condition is satisfied with a frequency of 10 Hz. If the condition is satisfied the execution of the language expression will be interrupted."

>  Can I understand a condition to be equivalent to a failure state, that wants to be avoided?
kind of? i mean it depends on the context. for example in suturo, they used monitoring to detect when a plate was in the correct position when placing it in the dishwasher. for this they monitored the force put onto the wrist of the hsr, and when it exceeded a certain value, it would mean that the plate is in the correct position, and the movement needs to be interrupted now. but i would not call that a failurestate.

src/pycram/language.py
sunava: done

LucaKro: you are right, i removed it

src/pycram/failure_handling.py
```diff
@@ -1,8 +1,11 @@
 from .designator import DesignatorDescription
 from .plan_failures import PlanFailure
+from threading import Lock
+from typing import Union
```
tomsch420, line 4: Import from typing_extension instead, names stay the same, python version compatibility increases

src/pycram/failure_handling.py
```diff
@@ -79,5 +82,98 @@ def perform(self):
...
 
+class RetryMonitor(FailureHandling):
+    """
+    A subclass of FailureHandling that implements a retry mechanism that works with a Monitor.
+
+    This class represents a specific failure handling strategy that allows us to retry a demo that is
+    being monitored, in case that monitoring condition is triggered.
+
+    Attributes:
+        max_tries (int): The maximum number of attempts to retry the action.
```
tomsch420, line 93: docstring for classmembers should be at the definition of the member. Check https://github.com/cram2/pycram/blob/d4a589d05d78807e0848ffab340fb0a6905b2734/src/pycram/designators/action_designator.py#L611 for an example. Furthermore use ReStructuredText as docstring formatting tool. For that you can go to PyCharm/Settings/Tools/Python Integrated Tools

src/pycram/failure_handling.py
```diff
@@ -79,5 +82,98 @@ def perform(self):
...
+        recovery (dict): A dictionary that maps exception types to recovery actions
+
+    Inherits:
+        All attributes and methods from the FailureHandling class.
```
tomsch420, line 97: Inherits and overrides are unnescessary since they are automatically described by sphinx

src/pycram/failure_handling.py
```diff
@@ -79,5 +82,98 @@ def perform(self):
...
+        perform(): Implements the retry logic.
+    """
+
+    def __init__(self, designator_description: Monitor, max_tries: int = 3, recovery: dict = None):
+        """
+        Initializes a new instance of the Retry class.
+
+        Args:
+            designator_description (DesignatorDescription): The description or context
+            of the task or process for which the retry mechanism is being set up.
```
tomsch420, line 109: RST format

src/pycram/failure_handling.py
```diff
@@ -79,5 +82,98 @@ def perform(self):
...
+                raise ValueError(
+                    "Recovery must be a dictionary with exception types as keys and Language instances as values.")
+            for key, value in recovery.items():
+                if not issubclass(key, BaseException):
+                    raise TypeError("Keys in the recovery dictionary must be exception types.")
+                if not isinstance(value, Language):
+                    raise TypeError("Values in the recovery dictionary must be instances of the Language class.")
+            self.recovery = recovery
+
+    def perform(self):
```
tomsch420, line 129: type annotation. From the docstring it should be  def perform(self) -> Tuple[bool, List[Status? (idk the inner type here. If not known, take Any)]]

src/pycram/failure_handling.py
```diff
@@ -79,5 +82,98 @@ def perform(self):
...
+        Implementation of the retry mechanism.
+
+        This method attempts to perform the Monitor + plan specified in the designator_description.
+        If the action fails, it is retried up to max_tries times. If all attempts fail,
+        the last exception is raised. In every loop, we need to clear the kill_event, and set all
+        relevant 'interrupted' variables to False, to make sure the Monitor and plan are executed
+        properly again
+
+        Raises:
+            PlanFailure: If all retry attempts fail.
```
tomsch420, line 140: RST format

src/pycram/failure_handling.py
```diff
@@ -79,5 +82,98 @@ def perform(self):
...
+
+        Returns:
+            The state of the execution performed, as well as a flattened list of the results, in the correct order
+        """
 
+        def reset_interrupted(child):
+            if hasattr(child, "interrupted"):
```
tomsch420, line 147: this is a very unsafe way of doing things. Why is child.interrupted not guaranteed?

src/pycram/failure_handling.py
```diff
@@ -79,5 +82,98 @@ def perform(self):
...
+                child.interrupted = False
+            for sub_child in getattr(child, "children", []):
```
tomsch420, line 149: Same as above

src/pycram/failure_handling.py
```diff
@@ -79,5 +82,98 @@ def perform(self):
...
+                try:
+                    status, res = self.designator_description.perform()
+                    break
+                except PlanFailure as e:
+                    tries += 1
+                    if tries >= self.max_tries:
+                        raise e
+                    exception_type = type(e)
+                    if exception_type in self.recovery:
+                        self.recovery[exception_type].perform()
```
tomsch420, line 178: thats pretty cool

src/pycram/language.py
```diff
@@ -260,34 +261,49 @@ def __init__(self, condition: Union[Callable, Fluent] = None):
         """
         super().__init__(None, None)
         self.kill_event = threading.Event()
+        self.exception_queue = queue.Queue()
```
tomsch420, line 264: Are you sure this shouldbe a queue? Python Queue does not mean queue, but synchronized queue. queue.dequeue is the asynchronous way i think.

src/pycram/language.py
```diff
@@ -260,34 +261,49 @@ def __init__(self, condition: Union[Callable, Fluent] = None):
...
-            if self.kill_event.is_set():
-                return
-            for child in self.children:
-                child.interrupt()
+            while not self.kill_event.is_set():
+                try:
+                    cond = self.condition.get_value()
+                    if cond:
+                        for child in self.children:
+                            if hasattr(child, 'interrupt'):
```
tomsch420, line 285: unsafe as above

tomsch420 (changes requested):

src/pycram/failure_handling.py
LucaKro: Oh i overlooked that one, should be fixed as soon as Vanessa merges my PR

src/pycram/failure_handling.py
LucaKro: Did the RST adjustements

src/pycram/failure_handling.py
LucaKro: i wasnt aware that its unsafe, removed it as it should not be needed anymore. where needed im using try-except now

src/pycram/language.py
LucaKro: the children are running in different threads, and as i understand queue.Queue is threadsafe, and synchronized queue makes sense here? I could be wrong though, input appreciated.
Also i looked up queue.dequeue, but only found queue.deque, but that does not seem to be the asynchronous way.

src/pycram/failure_handling.py
tomsch420: Hm, try except is not the solution i was looking for. If you expect the child to be of a certain type that has an interrupted attribute you should specify that in the typing and can expect that every child has it. Then just access the attribute. If someone passed in a wrong type he would get a yellow marking of the python linter and if executed anyway an attribute error that. Is that not intended?

src/pycram/language.py
tomsch420: Cool, then its correct.

src/pycram/failure_handling.py
LucaKro: Well the problem is that the type here may for example be a "Code()" language object, but due to python limitations, as i understand it, there are no ways to properly interrupt these code blocks. Thats why currently, the interrupt method of Code() raises a NotImplementedError. Also, this should never be called by hand, only through the Monitor, which can only do monitoring for Language objects. and these objects either have a interrupt method or it raises a NotImplementedError

src/pycram/failure_handling.py
sunava: done

src/pycram/language.py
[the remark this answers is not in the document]
LucaKro: done

src/pycram/failure_handling.py
LucaKro: done

src/pycram/failure_handling.py
sunava: @tomsch420 pls answer

tomsch420 (approved):

PR #180, opened 2024-08-13
[SemanticCostmap] applied margin

sunava:
I use this Costmap to generate placement poses for my complete pick-and-place demo on the real robot. The outer edges are reduced because the robot requires some clearance, especially when dealing with structures like shelves where the outer areas might be restricted. Since the standard semantic costmap is no longer in use, perhaps we should consider renaming this to something like placeSemanticCostmap, @Tigul. In the attached image, you can see an example for a countertop. Please note that I work with TF frames oriented according to the direction we agreed upon within the Suturo project.
Screenshot from 2024-08-13 12-17-33

### DIFF
--- a/src/pycram/costmaps.py
+++ b/src/pycram/costmaps.py
@@ -812,7 +812,7 @@ class SemanticCostmap(Costmap):
             self.map = self.map[:, max(0, non_zero_cols[0] - left_trim):non_zero_cols[-1] + 1 + right_trim]
 
 
-def get_aabb_for_link(self) -> AxisAlignedBoundingBox:
+    def get_aabb_for_link(self) -> AxisAlignedBoundingBox:
         """
         Returns the axis aligned bounding box (AABB) of the link provided when creating this costmap. To try and let the
         AABB as close to the actual object as possible, the Object will be rotated such that the link will be in the
--- a/src/pycram/costmaps.py
+++ b/src/pycram/costmaps.py
@@ -773,7 +773,7 @@ class SemanticCostmap(Costmap):
         self.width: int = 0
         self.map: np.ndarray = []
         self.generate_map()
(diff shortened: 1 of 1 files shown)


sunava:
for example uses:
 sem = SemanticCostmap(apartment_desig.resolve().world_object, "island_countertop")

sem.visualize()

```

File: src/pycram/costmaps.py [hunk 1]
```diff
@@ -772,19 +773,45 @@ def __init__(self, object, urdf_link_name, size=100, resolution=0.02, world=None
...
+        # Apply margin from one side (e.g., only the left)
+        if margin < self.width:
+            self.map[:, :margin] = 1
+            # # Right margin
+            self.map[:, -margin:] = 1
+
+        # Invert the map values: 0s become 1s, and everything else becomes 0
+        self.map = 1 - self.map
+
+        # Trim the left and right sides, only keep points before the 0s
```
Tigul, line 807: Why trim here when the margin is already applied above?

src/pycram/costmaps.py
```diff
@@ -753,7 +753,7 @@ class SemanticCostmap(Costmap):
     table surface.
     """
 
-    def __init__(self, object, urdf_link_name, size=100, resolution=0.02, world=None):
+    def __init__(self, object, urdf_link_name, size=100, resolution=0.02, world=None, margin_cm=0.2):
```
tomsch420, line 756: Something like margin_cm should just be margin: Unit. This way people can write whatever unit they want there. Consider using some metric package here like https://github.com/hgrecco/pint

tomsch420 (changes requested):

Tigul:
Since there will be another implementation soon I will close this

PR #181, opened 2024-08-13
[bullet world] added few functiontions - Add Environment Reference and Rigid Box Functionality

sunava:
The addition of the environment reference enhances the system's ability to handle different world configurations, especially in complex environments.
The new add_rigid_box() method is intended to streamline the creation of visual and collision objects in simulations, supporting more flexible environment interaction scenarios.
Please review these changes and provide feedback. Should we consider renaming the add_rigid_box() method for consistency with the other object-adding methods?


sunava:
we have new functions for that closing pr

PR #185, opened 2024-08-18
[specialized action] added our robocup action (not all yet but ok)

sunava:
As discussed in the meeting we adding this as specialized actions for now, and will see what we will merge into main actions


sunava:
ill merge them differently

PR #166, opened 2024-06-10
OntologyManager: add reason() + test

Tadinu:
Following #141, this PRs add reasoning APIs to `OntologyManager`, specifically to allow dynamic SWRL rules to be created for instances of an ontology concept class.
* Add `reason()`: reason on ontology concepts and their relations of a specific ontology world. Calling this will generate deducted facts that are updated to contents of property instances.
* Basic rules of binary relation (reflexivity, symmetry, transitivity) are introduced. NOTE: These just act as basic examples of rule definitions and more emphasize on the dynamic creation aspect. Same effects could be also achieved but incur creating new property classes, subclassing `ReflexiveProperty, SymmetricProperty, TransitiveProperty`
* Fix a bug causing reasoning to fail, due to missing required specific world as domain of ontology concepts for reasoning.
-> `LoadOntology()` now by default load from an IRI to the owlready2-provided `default_world`, which is also by default used for 
`sync_reasoner_pellet()/sync_reasoner_hermit()`.
-> Alternatively, `create_ontology_world()` is added in case of loading ontologies into a custom persistent world & reasoning on it specifically.
* `test_ontology.py`: Add `test_reasoning()`, also do post-cleaning up of `owl/sql, sql-journal` files generated during the test.
* `create_ontology_concept_class()/create_ontology_property_class()`: add optional `ontology` param to allow creating them into a specific ontology namespace, aside from the default `main_ontology`.
* `create_ontology_triple_classes()`:
  + (1) `predicate_name -> predicate_class_name`, `inverse_predicate_name -> inverse_predicate_class_name` for clarity
  + (2) add `predicate_python_attribute_name, inverse_predicate_python_attribute_name`, by default taking values from (1) if not  explicitly specified, for consistency
  + Make `ontology_object_parent_class` type hint as `Thing | object` to support such an `owlready2.FunctionalProperty` relation between `owlready2.Thing` and pure python types
* Docker/CI: add & verify Java runtime installation as required by the reasoner
https://owlready2.readthedocs.io/en/latest/reasoning.html#configuration

PR #141, opened 2024-04-04
Add owlready2-based OntologyConcept

duc89:
This PR adds API that enables designators to link themselves with concepts defined in ontologies, specifically ones of OWL format, utilizing **owlready2**.

-> This will enable robot motion plan to receive abstract ontology concepts as inputs, which will resolve themselves to specific physical objects in world, upon certain conditions/constraints from user scenario requirements.

Main updates are:
* Add `ontology.py` with `OntologyManager, providing various queries into ontology classes and properties, as well as relations among them.
* Add `ontology_common.py` with `OntologyConceptHolder` wrapping ontology_concept as an `owlready2.Thing` to avoid inheritance with added attributes that cannot be saved into an OWL file and reused later upon that file reloading.

Notably, 
  - Ontology triple classes can be also created dynamically with predicate relations set up from facts discovered in experimentation/robot training.
  - Those classes can also optionally inherit from ones existing in loaded ontologies.

* `DesignatorDescription, ObjectDesignatorDescription, ~ActionDesignatorDescription`: add `onto_concept_holders` A list of holders ontology concepts that the designator type is semantically categorized as or associated with

* Add `ontology.ipynb`, tutorial featuring basic usages of `OntologyManager` with examples used in robot motion plan
* `bullet_world_testcase.py, test_action_designator.py`: Add tests for ontology concepts loaded from SOMA

* `requirements.txt` add `owlready2`

@AbdelrhmanBassiouny @tomsch420

PR #188, opened 2024-08-20
Catch remote ontology loading error by owlready2

Tadinu:
Following #141 #166, this PR resolves issues related to ontology loading that obstruct `test_action_designator`
* `OntologyManager`: 
   * Catch ontology loading error thrown from owlready2 API, particularly for case of downloading a remote one (eg upon Internet 
connection glitch) -> to prevent unwanted disruption to the caller.
   * Group `main_ontology_world, main_ontology` creation into `create_main_ontology_world(), create_main_ontology()` to be also invokable on an alternative ontology IRI, given an loading fail from the initial one.
* `test_action_designator`: rem invalid check on `ontology_concept_holders`, which are only available in case of loading from SOMA, and also in fact more related to SOMA ontology contents themselves instead.

* `test_ontology`: defer to loading from local IRI in case of failed loading from the remote


Tadinu:
@tomsch420 @AbdelrhmanBassiouny @Tigul for review

tomsch420 (approved):

PR #190, opened 2024-08-21
Multiverse mjcf

AbdelrhmanBassiouny:
This depends on #177 

This adds the ability to use both MJCF and URDF for file descriptions with the new mjcf description class.

Also, it allows the creation of MJCF files from mesh files, similar to creating URDF from mesh files, however, the inertia and center of mass are automatically calculated.

This is done using MultiverseParser API to use this feature the Multiverse library is required.

Also added a description resolver to decide which description to use based on extension or an already given description like the GenericDescription from @sunava as it does not have a path or an extension in this case.


AbdelrhmanBassiouny:
The conflicts will be solved if #175 , #176 , and #177 are merged

PR #195, opened 2024-09-03
First draft for making semantic costmaps more flexible.

tomsch420:
This PR adds the 2D product algebra as a means to express locations for pycram.


demos/pycram_bullet_world_demo/demo.py
Tigul, line 1: I think this is wrongly commited

hawkina:
looks good, tests ran through. Merging...

PR #197, opened 2024-09-04
Fix missing import and error in tf_broadcaster

J-Schaefer:
Ran into an error here. Since the property projection_namespace is now of type ExecutionType it cannot be concatenated with a string anymore. So we can take the name for that.

### DIFF
--- a/src/pycram/ros/tf_broadcaster.py
+++ b/src/pycram/ros/tf_broadcaster.py
@@ -5,6 +5,7 @@ import atexit
 
 from ..datastructures.pose import Pose
 from ..datastructures.world import World
+from ..datastructures.enums import ExecutionType
 from tf2_msgs.msg import TFMessage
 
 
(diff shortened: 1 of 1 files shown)


src/pycram/ros/tf_broadcaster.py
```diff
@@ -78,8 +79,8 @@ def _publish_pose(self, child_frame_id: str, pose: Pose, static=False) -> None:
         frame_id = pose.frame
         if frame_id != child_frame_id:
             tf_stamped = pose.to_transform(child_frame_id)
-            tf_stamped.frame = self.projection_namespace + "/" + tf_stamped.frame
-            tf_stamped.child_frame_id = self.projection_namespace + "/" + tf_stamped.child_frame_id
+            tf_stamped.frame = self.projection_namespace.name + "/" + tf_stamped.frame
```
tomsch420, line 82: isnt os.path.join nicer here?

J-Schaefer: Not really the scope of this PR, since this is only fixing a bug, but tried it nevertheless. `os.path.join` behaves differently depending on the platform. No problem if we are always on Linux with this. I could add `"/".join(...)` though.

tomsch420: Bug fixes are also required to be as clean as possible.
The question was more If for another os than Linux ghe rostopic still has "/" as seperator or If eg on the seperator will be "\"

tomsch420: *Backslash

PR #191, opened 2024-08-23
Ontology to pycram objects

mpomarlan:
Work in progress, but started a pull request to draw attention to this.

TODOs:

1. SERIOUS: get SOMA_DFL and/or SOMA_DFL_module online so the onto loader can get to them without needing to depend on the dfl python package. Currently, these would be accessible only on a github tree.
2. Serious: test to remove silly typos
3. Good to have (but not threatening to functionality, at least in limited use cases): Integration with rest of onto manager. E.g., I prefer to have an iri2Concept map to switch between string names and the actual OWLReady2 generated classes. What other functions in the onto manager may impact this mapping?
4. Good to have (eventually): query complex concept expressions. This will require integrating a different reasoner in there (Konclude) or only use modules of SOMA_DFL so that OWLReady2's HermiT doesn't die.
5. Good to have (eventually): a more elegant map between pycram object types and ontology classes.

PR #198, opened 2024-09-05
[OntologyManager] Use in-memory SQL backend by default

Tadinu:
Related to #141 #166 #191, this PR switches the default SQL-backend to `in-memory` & add general catch of `sqlite3.Error` from owlready2 accessing to SQL backend as a disk-file, to avoid disrupting Designator functionality. 

Specifically,
* `OntologyManager.create_ontology_world()` -> `owlready2.World.set_backend()`, where sqlite3 DB access commands are invoked. Errors may happen upon a corrupt db or concurrent accesses, or a typical `OperationalError` would be raised amid debug process, which if sometimes prematurely ended (eg by pause/stop-debug in IDE), could result in a locked-db file on disk.
Such errors would be treated in another PR, or even better by `owlready2` itself.
https://stackoverflow.com/a/2741015/4767815

Also, A quick resolution is deleting the SQL file on disk.
SQL backend essentially acts like a cache that is only required during program run, one if not-existing is automatically re-generated afresh from contents from the `.owl`, which together with all ontology data (including dynamically created concepts) can be saved anyway at program exit.

* `OntologyManager`: add `main_ontology_sql_backend` attr, allowing users to specifically set path to an SQL backend file (no need to already exist), otherwise it will be `in-memory`.
* `OntologyManager.create_main_ontology_world()`: catch `sqlite3.Error`
* `OntologyManager.load_ontology()`: in case `ontology_iri` is a local relative path & non-existing, make sure the file creating location is under `ontology_search_path`, which is settable by users
* `test_ontology`: only test sql_backend being saved to file in case of non-in-memory backend.

### DIFF
--- a/src/pycram/ontology/ontology.py
+++ b/src/pycram/ontology/ontology.py
@@ -4,6 +4,7 @@ import inspect
 import itertools
 import logging
 import os.path
+import sqlite3
 
 from pathlib import Path
 from typing import Callable, Dict, List, Optional, Type, Tuple, Union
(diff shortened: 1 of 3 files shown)


Tadinu:
@tomsch420 @Tigul as reviewers

tomsch420 (approved):

PR #203, opened 2024-09-19
Improve installation process

rhaschke:
See individual commit messages for details

### DIFF
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -107,6 +107,18 @@ Then install the Python packages in the requirements.txt file
     sudo pip3 install -r requirements.txt
     sudo pip3 install -r src/neem_interface_python/requirements.txt
 
+This installs the packages into ``/usr/local/lib``. If you prefer to not clutter your system-wide python installation,
+you can also install the packages into the catkin workspace as follows:
+
+.. code-block:: shell
+
+    # install packages into catkin workspace instead of ~/.local
+    export PYTHONUSERBASE=~/workspace/ros/devel
+    # don't install packages that are available in system
+    export PIP_IGNORE_INSTALLED=0
+
+    pip3 install -r requirements.txt
+    pip3 install -r src/neem_interface_python/requirements.txt
 
 Building your ROS workspace
 ===========================
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -105,7 +105,6 @@ Then install the Python packages in the requirements.txt file
 .. code-block:: shell
 
     sudo pip3 install -r requirements.txt
-    sudo pip3 install -r src/neem_interface_python/requirements.txt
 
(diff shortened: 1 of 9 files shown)


doc/source/installation.rst
```diff
@@ -107,6 +107,18 @@ Then install the Python packages in the requirements.txt file
...
+
+.. code-block:: shell
+
+    # install packages into catkin workspace instead of ~/.local
+    export PYTHONUSERBASE=~/workspace/ros/devel
+    # don't install packages that are available in system
+    export PIP_IGNORE_INSTALLED=0
+
+    pip3 install -r requirements.txt
+    pip3 install -r src/neem_interface_python/requirements.txt
```
Tigul, line 121: Please remove this line, since the neem_interface_python is not part of the repo anymore.

doc/source/remarks.rst
```diff
@@ -33,14 +33,3 @@ Dirty Terminals
 
 If your terminal gets polluted by PyBullet complaining about incomplete URDF descriptions, you need to first fix your
 URDF files by inserting the missing tags and second delete the `resources/cached` folder.
-
-Missing pr2_arm_kinematics
-==========================
```
Tigul, line 38: Why did you remove this part?

doc/source/troubleshooting.rst
```diff
@@ -108,32 +108,3 @@ real_robot environments. This is also explained in the `Action Designator Exampl
 
    with simulated_robot:
       NavigateAction([Pose()]).resolve().perform()
-
-Missing pr2_arm_kinematics
-==========================
```
Tigul, line 113: Same here, why remove this?

Tigul (changes requested):

doc/source/installation.rst
rhaschke: Ok. I didn't use that and thus didn't notice.

doc/source/remarks.rst
rhaschke: The arm kinematics error was caused by a missing dependency to `moveit_kinematics`. As I added that dependency in ef50689f64ff05106015a3375b037c4037c0b1c4, the error should be gone now.

Tigul: Yeah but that error occurred for some people randomly after some time of usage, so I would like to keep this for documentation at least in the troubleshooting section.

rhaschke: As stated in the removed doc, the package `moveit_kinematics` was auto-removed by apt after some time, which happened because the dependency to `moveit_kinematics` was not declared. As it is now, the package will not be removed accidentally anymore. Trust me ;-)
If you insist, I will revert that part.

Tigul: Ok, then I will trust you on that :)

PR #205, opened 2024-10-04
Ros decoupling

Tigul:
# Description 
This PR creates  a new module which abstracts all calls to ROS, this is in preparation for the ROS2 upgrade.


src/pycram/worlds/bullet_world.py
```diff
@@ -35,10 +33,6 @@ class is the main interface to the Bullet Physics Engine and should be used to s
     manipulate the Bullet World.
     """
 
-    # Check is for sphinx autoAPI to be able to work in a CI workflow
-    if rosgraph.is_master_online():  # and "/pycram" not in rosnode.get_node_names():
```
tomsch420, line 39: can this be safely removed?

tomsch420 (changes requested):

src/pycram/worlds/bullet_world.py
Tigul: Yes, I moved this to the __init__.py of the ros package. So it should be called anytime something wants to interact with ROS.

PR #206, opened 2024-10-09
Mesh and object

AbdelrhmanBassiouny:
Added some extra functionality to use information from the meshes and adjust the meshes before they are cached if needed.

Added the option of getting the axis_aligned_bounding_box from mesh and link geometry in general instead of only from simulator, this adds more flexibility and does not require the simulator always to provide this functionality.

Also now there is rotated_bounding_box, and convex hull.

Added the option to provide callbacks for object addition and removal to be used by the users if needed.


src/pycram/cache_manager.py
```diff
@@ -50,7 +52,8 @@ def delete_cache_dir(self):
 
     def update_cache_dir_with_object(self, path: str, ignore_cached_files: bool,
                                      object_description: 'ObjectDescription', object_name: str,
-                                     scale_mesh: Optional[float] = None) -> str:
+                                     scale_mesh: Optional[float] = None,
+                                     mesh_transform: Optional['Transform'] = None) -> str:
```
Tigul, line 56: You don't need to put Transform in quotation marks if you add the import:
from __future__ import annotations

src/pycram/datastructures/dataclasses.py
```diff
@@ -99,15 +100,24 @@ class AxisAlignedBoundingBox:
...
+        return [[point.x, point.y, point.z] for point in self.get_points()]
 
-        :param min_point: The minimum point
-        :param max_point: The maximum point
+    def get_points(self) -> List[Point]:
         """
-        return cls(min_point[0], min_point[1], min_point[2], max_point[0], max_point[1], max_point[2])
+        :return: The points of the bounding box as a list of Point instances.
+        """
+        return [Point(self.min_x, self.min_y, self.min_z),
```
Tigul, line 113: Is this intended like this? and why are there so many points?

src/pycram/description.py
```diff
@@ -219,6 +220,53 @@ def __init__(self, _id: int, link_description: LinkDescription, obj: Object):
         self._current_pose: Optional[Pose] = None
         self.update_pose()
 
+    def get_bounding_box(self, rotated: bool = False) -> Union[AxisAlignedBoundingBox, RotatedBoundingBox]:
+        """
+        :param rotated: If True, return the rotated bounding box, otherwise the axis-aligned bounding box.
+        :return: The axis-aligned or rotated bounding box of a link. First try to get it from the simulator, if not,
+         then calculate it depending on the type of the link geometry.
```
Tigul, line 227: Add return to doc string

src/pycram/cache_manager.py
AbdelrhmanBassiouny: oh that's nicee thanks

src/pycram/datastructures/dataclasses.py
AbdelrhmanBassiouny: Yes these are the 8 corners of the bounding box.

src/pycram/description.py
AbdelrhmanBassiouny: What do you mean? there is already a :return: in the doc string.

src/pycram/datastructures/dataclasses.py
```diff
@@ -99,53 +108,150 @@ class AxisAlignedBoundingBox:
...
+        Set the axis-aligned bounding box from a minimum and maximum point.
+
+        :param min_point: The minimum point
+        :param max_point: The maximum point
+        """
+        return cls(min_point[0], min_point[1], min_point[2], max_point[0], max_point[1], max_point[2])
+
+
+@dataclass
+class RotatedBoundingBox(BoundingBox):
```
tomsch420, line 206: Why dont use a polytope here

AbdelrhmanBassiouny: We can add the convex hull version which is available directly from trimesh, but this does not remove the need for a rotated shape, because the rotation here is meant to rotate the shape to the current pose of the link, this rotation will also be needed in a polytope or a convex hull as we only get the points relative to the object origin and are aligned with world frame not with object current frame.

AbdelrhmanBassiouny: I have added a method to get convex hull

src/pycram/costmaps.py
```diff
@@ -785,7 +785,7 @@ def get_aabb_for_link(self) -> AxisAlignedBoundingBox:
             link_pose_trans = self.link.transform
             inverse_trans = link_pose_trans.invert()
             prospection_object.set_orientation(inverse_trans.to_pose())
-            return self.link.get_axis_aligned_bounding_box()
+            return self.link.get_bounding_box()
```
Tigul, line 788: Wouldn't this mean that the bounding box is not necessarily axis aligned?

src/pycram/datastructures/dataclasses.py
```diff
@@ -26,6 +27,14 @@ def get_point_as_list(point: Point) -> List[float]:
     return [point.x, point.y, point.z]
 
 
+def get_list_from_points(points: List[Point]) -> List[List[float]]:
```
Tigul, line 30: No documentation

src/pycram/datastructures/dataclasses.py
```diff
@@ -99,53 +108,150 @@ class AxisAlignedBoundingBox:
...
+
+        :param axis_aligned_bounding_box: The axis-aligned bounding box.
+        :param transform: The transformation.
+        """
+        return cls(axis_aligned_bounding_box.min_x, axis_aligned_bounding_box.min_y, axis_aligned_bounding_box.min_z,
+                   axis_aligned_bounding_box.max_x, axis_aligned_bounding_box.max_y, axis_aligned_bounding_box.max_z,
+                   transform)
+
+    def get_corners(self, transform: Optional[Transform] = None) -> List[List[float]]:
+        """
```
Tigul, line 243: No documentation

src/pycram/datastructures/world.py
```diff
@@ -807,8 +857,6 @@ def set_link_color(self, link: Link, rgba_color: Color):
     @abstractmethod
     def get_link_color(self, link: Link) -> Color:
         """
```
Tigul, line 859: Remonde doc?

src/pycram/datastructures/world.py
```diff
@@ -817,32 +865,45 @@ def get_link_color(self, link: Link) -> Color:
...
         :param obj: The object
-        :return: A dictionary with link names as keys and a Color object for each link as value.
+        :return: The RGBA colors of each link in the object as a dictionary from link name to rgba_color.
         """
         pass
 
     @abstractmethod
     def get_object_axis_aligned_bounding_box(self, obj: Object) -> AxisAlignedBoundingBox:
         """
-        Return the axis aligned bounding box of this object. The return of this method are two points in
```
Tigul, line 830: Doc?

src/pycram/description.py
Tigul: Good question

src/pycram/world_concepts/world_object.py
```diff
@@ -487,7 +501,7 @@ def get_link_axis_aligned_bounding_box(self, link_name: str) -> AxisAlignedBound
         :param link_name: The name of the link.
         :return: The axis aligned bounding box of the link.
         """
-        return self.links[link_name].get_axis_aligned_bounding_box()
+        return self.links[link_name].get_bounding_box()
```
Tigul, line 504: This could Return a bounding box that is Not Anis aligned which is confusing given the Method Name

tomsch420 (approved):

src/pycram/datastructures/dataclasses.py
AbdelrhmanBassiouny: Done

src/pycram/datastructures/dataclasses.py
AbdelrhmanBassiouny: Done

src/pycram/costmaps.py
AbdelrhmanBassiouny: yes you are right I removed it and added two methods one for axis_ligned and one for rotated

src/pycram/world_concepts/world_object.py
AbdelrhmanBassiouny: fixed it

Tigul (approved):

PR #212, opened 2024-10-18
Notebook test ci

Tigul:
# Description 
Added a CI pipeline that automacillay runs all example notebooks and checks whether they are working or not 

Also fixed a lot of issues with outdated code in the notebooks


.github/workflows/notebook-test-ci.yml
```diff
@@ -0,0 +1,83 @@
...
+          pip3 install jupytext treon
+
+      # ----------------------------------------------------------------------------------------------------------------
+      - name: Convert Notebooks
+        run: |
+          cd /opt/ros/overlay_ws/src/pycram/examples
+          rm -rf tmp
+          mkdir tmp
+          jupytext --to notebook *.md
+          mv *.ipynb tmp && cd tmp
```
tomsch420, line 72: shouldnt this be part of the jb build pipeline?

src/pycram/worlds/bullet_world.py
```diff
@@ -448,6 +449,9 @@ def __init__(self, world: World, mode: WorldMode):
         self.world = world
         self.mode: WorldMode = mode
 
+        if not "DISPLAY" in os.environ:
```
tomsch420, line 452: document this. Otherwise some user with some weird display setup may complain about not having a bullet world and this will be hell to figure out why

tomsch420 (changes requested):

.github/workflows/notebook-test-ci.yml
Tigul: Do you mean the pipeline for the documentation?

tomsch420: yes

Tigul: No this is specifically so that treon can do the testing for the notebooks

tomsch420 (approved):

PR #219, opened 2024-11-14
[doc] Added doc about costmaps

Tigul:
Added some more documentation about costmaps which explains them in a bit more detail


doc/images/gaussian_costmap.png
tomsch420, line 1: that is pretty ugly ngl

tomsch420 (approved):

PR #220, opened 2024-11-19
Better Probabilistic Costmaps

tomsch420:
Analogously to  PR#215 this is a proper implementation of costmaps using well defined distributions.

### DIFF
--- a/src/pycram/probabilistic_costmap.py
+++ b/src/pycram/probabilistic_costmap.py
@@ -1,18 +1,25 @@
 from enum import Enum, auto
+from functools import cached_property
 
 import numpy as np
 import tf
+from geometry_msgs.msg import Point
+from std_msgs.msg import ColorRGBA
 from tf.transformations import quaternion_from_matrix
 from random_events.interval import closed, closed_open
 from typing_extensions import Optional, Type
+from visualization_msgs.msg import Marker, MarkerArray
 
 from . import World
 from .costmaps import Costmap, OccupancyCostmap, VisibilityCostmap
+import matplotlib.colorbar
 from .datastructures.pose import Pose
+from .ros.data_types import Duration
+from .ros.publisher import create_publisher
 from .units import meter, centimeter
 
 from pint import Unit, Quantity
-from probabilistic_model.probabilistic_circuit.nx.helper import uniform_measure_of_event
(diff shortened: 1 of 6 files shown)


src/pycram/probabilistic_costmap.py
```diff
@@ -0,0 +1,166 @@
...
+import numpy as np
+import tf
+from geometry_msgs.msg import Point
+from std_msgs.msg import ColorRGBA
+from tf.transformations import quaternion_from_matrix
+from random_events.interval import closed, closed_open
+from typing_extensions import Optional, Type
+from visualization_msgs.msg import Marker, MarkerArray
+
+from . import World
```
Tigul, line 13: Don't import World like this, this could easily lead to circular imports

src/pycram/probabilistic_costmap.py
```diff
@@ -0,0 +1,166 @@
...
+        self.origin = origin
+        self.size = size
+
+        # calculate the number of cells per axis
+        number_of_cells = int(np.sqrt(max_cells))
+        resolution = self.size.to(meter) / number_of_cells
+
+        if costmap_type == OccupancyCostmap:
+            self.costmap = OccupancyCostmap(
+                origin=self.origin,
```
Tigul, line 84: Initializing the costmap with static arguments is very inflexible. It could be a better idea to take the parameter for the costmap as kwargs

src/pycram/probabilistic_costmap.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+        Create a probabilistic circuit from the costmap.
+        """
+        # self.distribution = fully_factorized([self.x, self.y], {self.x: self.origin.position.x,
+        #                                                     self.y: self.origin.position.y},
+        #                                      {self.x: 1, self.y: 1})
+        # self.distribution, _ = self.distribution.conditional(self.create_event_from_map())
+        self.distribution = uniform_measure_of_event(self.create_event_from_map())
+
+    def sample_to_pose(self, sample: np.ndarray) -> Pose:
```
Tigul, line 124: There is a PoseGenerator class for this which also allows different orientation generators. That would be better to use

src/pycram/units.py
Tigul, line 1: Nice idea

PR #230, opened 2024-11-27
[FTS] real robot tested on hsrb and donbot

sunava:
[fts]  Butterworth  filter for Force Torque Sensor 
- tested on hsr and donbot in real 
- test cases with mockup 

for understanding:
The `@patch` decorator in Python is used in unit testing to replace objects in your code with mock objects during the test. This allows you to control the behavior of dependencies and isolate the code under test. It is part of the `unittest.mock` module.
Here is a brief explanation of how it works:

- `@patch` replaces the specified object with a mock object.
- The mock object can be configured to return specific values or raise exceptions.
- After the test, the original object is restored.

Example usage:
```python
from unittest.mock import patch

@patch('module.ClassName')
def test_function(mock_class):
    # mock_class is a mock object replacing module.ClassName
    instance = mock_class.return_value
    instance.method.return_value = 'mocked value'

    # Your test code here

```

### DIFF
--- a/test/test_force_torque_sensor.py
+++ b/test/test_force_torque_sensor.py
@@ -11,8 +11,9 @@ class ForceTorqueSensorTestCase(unittest.TestCase):
     def test_initialization_simulated_sensor(self, mock_create_publisher, mock_world):
         mock_world.current_world.robot.joint_name_to_id = {'joint1': 1}
         sensor = ForceTorqueSensorSimulated('joint1')
-        self.assertEqual(sensor.joint_name, 'joint1')
+        self.assertEqual('joint1', 'joint1')
         self.assertIsNotNone(sensor.fts_pub)
+        sensor._stop_publishing()
 
     @patch('pycram.ros_utils.force_torque_sensor.World')
     @patch('pycram.ros_utils.force_torque_sensor.create_publisher')


src/pycram/ros/custom_filter.py
```diff
@@ -0,0 +1,29 @@
+from abc import abstractmethod
+
+from scipy.signal import butter, lfilter
+
+
+class CustomFilter:
+    """
+    Abstract class to ensure that every supported filter needs to implement the filter method
```
tomsch420, line 8: can the class name also be filter?

src/pycram/ros/custom_filter.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+
+    @abstractmethod
+    def filter(self, data):
```
tomsch420, line 12: docstring

src/pycram/ros/custom_filter.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Butterworth(CustomFilter):
+    """
+    Implementation for a Butterworth filter.
+    """
+
+    def __init__(self, order=4, cutoff=10, fs=60):
+        self.order = order
+        self.cutoff = cutoff
```
tomsch420, line 23: documentation of params

src/pycram/ros_utils/force_torque_sensor.py
```diff
@@ -77,3 +82,152 @@ def _stop_publishing(self) -> None:
...
+        Create a subscriber for the force-torque-sensor topic of a specified robot.
+
+        :param robot_name: Name of the robot
+        :param filter_config: Desired filter (default: Butterworth)
+        :param filter_order: Order of the filter. Declares the number of elements that delay the sampling
+        :param custom_topic: Declare a custom topic if the default topics do not fit
+        """
+
+        self.robot_name = robot_name
+        self.filter_config = filter_config
```
tomsch420, line 112: declare the variables as class variables and document them there

src/pycram/ros_utils/force_torque_sensor.py
```diff
@@ -77,3 +82,152 @@ def _stop_publishing(self) -> None:
...
+        self.init_data = True
+
+        self.whole_data = None
+        self.prev_values = None
+
+        self.order = filter_order
+
+        self.__setup()
+
+    def __setup(self):
```
tomsch420, line 127: double underscore methods usually have a meaning of "super private please do never access this method from any other file". Is this intended?

src/pycram/ros_utils/force_torque_sensor.py
```diff
@@ -77,3 +82,152 @@ def _stop_publishing(self) -> None:
...
+        filtered_data = self.__filter_data(data_compensated)
+
+        self.whole_data[self.unfiltered].append(data_compensated)
+        self.whole_data[self.filtered].append(filtered_data)
+
+        self.prev_values.append(data_compensated)
+        self.prev_values.pop(0)
+
+        if self.debug:
+            print(
```
tomsch420, line 160: use the debug logging here instead of printing

tomsch420 (changes requested):
Please also provide test cases

PR #234, opened 2024-12-02
Object dynamic state

AbdelrhmanBassiouny:
This depends on #227 
Created a super class for Object and Link called PhysicalBody, this hast common attributes for any physical body like pose, velocity, contact points.

Also changed ContactPoint to use bodies instead of links as datatype for returned bodies.

Also Everything now is a WorldEntity including PhysicalBody, and Wolrd, this unified the methods __eq__ and __hash__ so they are only implemented in WorldEntity except for World which overwrites them.

Also every WorldEntity has a parent_entity, so Link has parent as Object and Object has parent as world, while World has None.

In addition the ontology_concept class variable was moved from Object class to PhysicalBody class because now it makes more sense there since any PhysicalBody is a PhysicalObject by Default.

Added attributes for PhysicalBody like velocity, is_translating, is_rotating, is_moving but currently they are not set by anything, one way could be through querying the simulation but one could also set them by performing their own calculation and according to their own thresholds for that, they just are helpful to be inside physical body if someone is interested in its dynamic state.

Also Added a PhysicalBodyState that includes pose, velocity, is_translating, and is_rotating.


demos/pycram_multiverse_demo/demo_euROBIN_industrial_robotics.py
```diff
@@ -32,3 +32,54 @@
...
+
+
+if __name__ == '__main__':
+    root = logging.getLogger()
+    root.setLevel(logging.INFO)
+
+    world = Multiverse(simulation_name="ur5e_with_task_board")
+
+    # Load environment, robot and objects
+    rospack = RosPack()
```
Tigul, line 75: RosPack might break with ROS2, just as a heads up. Also I don't really understand what you are using this for?

src/pycram/datastructures/dataclasses.py
```diff
@@ -478,15 +479,78 @@ class State(ABC):
...
+    Dataclass for storing the state of a physical body.
+    """
+    pose: Pose
+    is_translating: bool
+    is_rotating: bool
+    velocity: List[float]
+    acceptable_pose_error: Tuple[float, float] = (0.001, 0.001)
+    acceptable_velocity_error: Tuple[float, float] = (0.001, 0.001)
+
+    def __eq__(self, other: 'PhysicalBodyState'):
```
Tigul, line 494: Import from __future__ import annotations and remove the quotation marks

src/pycram/datastructures/dataclasses.py
```diff
@@ -478,15 +479,78 @@ class State(ABC):
...
+        return (self.pose_is_almost_equal(other)
+                and self.is_translating == other.is_translating
+                and self.is_rotating == other.is_rotating
+                and self.velocity_is_almost_equal(other)
+                )
+
+    def pose_is_almost_equal(self, other: 'PhysicalBodyState') -> bool:
```
Tigul, line 501: Also here

src/pycram/datastructures/world.py
```diff
@@ -737,47 +700,66 @@ def perform_collision_detection(self) -> None:
         """
         pass
 
-    @abstractmethod
+    @deprecated("Use get_body_contact_points instead")
```
Tigul, line 703: Nice

src/pycram/datastructures/world_entity.py
```diff
@@ -110,10 +120,275 @@ def remove_saved_states(self) -> None:
...
+        pass
+
+    @property
+    def contact_points(self) -> ContactPointsList:
+        """
+        :return: The contact points of this body with other physical bodies.
+        """
+        return self.world.get_body_contact_points(self)
+
+    def get_contact_points_with_body(self, body: 'PhysicalBody') -> ContactPointsList:
```
Tigul, line 328: import from __future__ import annotations and remove the quotation marks

src/pycram/description.py
```diff
@@ -303,7 +294,7 @@ def get_mesh_filename(self, geometry: MeshVisualShape) -> str:
         else:
             raise LinkGeometryHasNoMesh(self.name, type(geometry).__name__)
 
-    def set_pose(self, pose: Pose) -> None:
+    def set_object_pose_given_link_pose(self, pose: Pose) -> None:
```
Tigul, line 297: That name sounds confusing

src/pycram/world_concepts/constraints.py
```diff
@@ -51,7 +51,7 @@ def set_child_link_pose(self):
         """
         Set the target pose of the child object to the current pose of the child object in the parent object frame.
         """
-        self.child_link.set_pose(self.get_child_link_target_pose())
+        self.child_link.set_object_pose_given_link_pose(self.get_child_link_target_pose())
```
Tigul, line 54: Also confusing name

demos/pycram_multiverse_demo/demo_euROBIN_industrial_robotics.py
AbdelrhmanBassiouny: oh, that shouldn't be there I removed it.

src/pycram/datastructures/dataclasses.py
AbdelrhmanBassiouny: done

src/pycram/datastructures/world.py
AbdelrhmanBassiouny: Thanks :D

src/pycram/datastructures/dataclasses.py
AbdelrhmanBassiouny: done

src/pycram/datastructures/world_entity.py
AbdelrhmanBassiouny: done

src/pycram/description.py
AbdelrhmanBassiouny: hmm the thing is the link pose shouldn't be settable, but you can move the object such that the link has this given pose, and Iw ould like users to know that what they are doing is actually moving the whole object when ding this

AbdelrhmanBassiouny: so there is the usual set_pose of the link but I did there was issue a warning that link poses are not settable.

Tigul: After thinking about this I think you are right

PR #235, opened 2024-12-02
[detectaction] Robokudo real robot

sunava:
# DetectAction Overview

The `DetectAction` now returns a list of all detected objects. It supports various detection techniques:

- **ALL**: `0`
- **HUMAN**: `1`
- **TYPES**: `2`
- **REGION**: `3`
- **HUMAN_ATTRIBUTES**: `4`
- **HUMAN_WAVING**: `5`

## Real Environment Behavior
Objects perceived in the real world are dynamically spawned into the environment.

### Simulation Behavior
- The `get_scene_objects` method serves as a mockup for the `ALL` detection type.
- Detection types like `HUMAN` and `HUMAN_ATTRIBUTES` are **not supported** in the simulation.

## Important Note
In the real Robokudo system, filtering for specific objects is **not supported** by default. While the detection pipeline is correct, if you are using it in a real environment, you must manually filter the objects in your demo.

## State Management
- The `state` parameter manages continuous perception.
- Default state: `start`.
- The perception can be stopped by setting `state=stop`.

---

## To-Do
@Tigul: Implement knowledge condition.
```python
# Example:
# self.knowledge_condition = VisibleProperty(self.object_designator_description)

### DIFF
--- a/src/pycram/datastructures/world.py
+++ b/src/pycram/datastructures/world.py
@@ -12,7 +12,7 @@ from geometry_msgs.msg import Point
 from typing_extensions import List, Optional, Dict, Tuple, Callable, TYPE_CHECKING, Union, Type
 
 import pycrap
-from pycrap import PhysicalObject, Floor, Apartment
+from pycrap import PhysicalObject, Floor, Apartment, Robot
 from ..cache_manager import CacheManager
 from ..config.world_conf import WorldConfig
 from ..datastructures.dataclasses import (Color, AxisAlignedBoundingBox, CollisionCallbacks,
@@ -353,10 +353,9 @@ class World(StateEntity, ABC):
 
     def get_scene_objects(self) -> List[Object]:
         """
-        Return all objects in the world except the robot, floor and apartment.
-        :return: A list of all objects in the world except the robot, floor and apartment.
+        :return: A list of all objects in the world except the robot, floor, and apartment.
         """
-        return [obj for obj in self.objects if obj not in {self.robot, Floor, Apartment}]
+        return [obj for obj in self.objects if obj.obj_type not in {Robot, Floor, Apartment}]
 
     def remove_visual_object(self, obj_id: int) -> bool:
         """
--- a/src/pycram/datastructures/world.py
+++ b/src/pycram/datastructures/world.py
@@ -12,7 +12,7 @@ from geometry_msgs.msg import Point
 from typing_extensions import List, Optional, Dict, Tuple, Callable, TYPE_CHECKING, Union, Type
 
 import pycrap
-from pycrap import PhysicalObject
+from pycrap import PhysicalObject, Floor, Apartment
 from ..cache_manager import CacheManager
 from ..config.world_conf import WorldConfig
 from ..datastructures.dataclasses import (Color, AxisAlignedBoundingBox, CollisionCallbacks,
@@ -350,6 +350,13 @@ class World(StateEntity, ABC):
         :return: The Object with the id 'id'.
         """
         return list(filter(lambda obj: obj.id == obj_id, self.objects))[0]
+
+    def get_scene_objects(self) -> List[Object]:
+        """
+        Return all objects in the world except the robot, floor and apartment.
+        :return: A list of all objects in the world except the robot, floor and apartment.
+        """
+        return [obj for obj in self.objects if obj not in {self.robot, Floor, Apartment}]
 
     def remove_visual_object(self, obj_id: int) -> bool:
         """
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

File: src/pycram/datastructures/world.py [hunk 1]
```diff
@@ -351,6 +351,13 @@ def get_object_by_id(self, obj_id: int) -> Object:
         """
         return list(filter(lambda obj: obj.id == obj_id, self.objects))[0]
 
+    def get_scene_objects(self) -> List[Object]:
+        """
+        Return all objects in the world except the robot, floor and apartment.
+        :return: A list of all objects in the world except the robot, floor and apartment.
+        """
+        return [obj for obj in self.objects if obj not in {self.robot, Floor, Apartment}]
```
Tigul, line 359: Floor and Apartment are PyCRAP types, filtering this way will not work as intended. You need to find the respective object by their type first. 
Also, please add a new line between the doc string and the :return:

src/pycram/process_modules/default_process_modules.py
```diff
@@ -1,9 +1,14 @@
 from threading import Lock
+from typing import List
```
Tigul, line 2: Please use typing_extensions instead

src/pycram/process_modules/default_process_modules.py
```diff
@@ -180,6 +216,94 @@ def _move_arm_tcp(target: Pose, robot: Object, arm: Arms) -> None:
...
+                except IndexError:
+                    pass
+                obj_color = None
+                try:
+                    obj_color = query_result.res[i].color[0]
+                except IndexError:
+                    pass
+
+                # Map color names to RGBA values
+                color_switch = {
```
Tigul, line 265: Move this to utils

src/pycram/process_modules/default_process_modules.py
```diff
@@ -180,6 +216,94 @@ def _move_arm_tcp(target: Pose, robot: Object, arm: Arms) -> None:
...
+                    color = Color(0, 0, 0, 1)
+
+                hsize = [obj_size.x / 2, obj_size.y / 2, obj_size.z / 2]
+
+                # Check if the object type is a subclass of the classes in the objects module (pycrap)
+                class_names = [name for name, obj in inspect.getmembers(objects, inspect.isclass)]
+
+                matching_classes = [class_name for class_name in class_names if obj_type in class_name]
+
+                obj_name = obj_type + "" + str(rospy.get_time())
```
Tigul, line 289: Please move the rospy.get_time function to pycram.ros

src/pycram/process_modules/pr2_process_modules.py
```diff
@@ -80,6 +81,7 @@ def _execute(self, desig: MoveGripperMotion):
             robot.set_joint_position(joint, state)
 
 
+@DeprecationWarning
 class Pr2Detecting(ProcessModule):
```
Tigul, line 85: You can delete both Detecting process modules since they are not used anymore but good idea with the DeprecationWarning

test/test_designator/test_action_designator.py
```diff
@@ -96,14 +96,14 @@ def test_look_at(self):
     def test_detect(self):
         self.kitchen.set_pose(Pose([10, 10, 0]))
         self.milk.set_pose(Pose([1.5, 0, 1.2]))
-        object_description = object_designator.ObjectDesignatorDescription(names=["milk"])
-        description = action_designator.DetectAction(object_description)
-        self.assertEqual(description.ground().object_designator.name, "milk")
+        object_description = object_designator.ObjectDesignatorDescription(types=[Milk])
```
Tigul, line 99: Please import the designator directly and use it instead of using the module

Tigul (changes requested):

src/pycram/datastructures/world.py
tomsch420: Furthermore, if you comment just describes the return value you can just put it behind the :return: so nothing getsduplicated

src/pycram/designators/action_designator.py
```diff
@@ -460,17 +461,32 @@ def plan(self) -> None:
...
+
+    If no object is found, an PerceptionObjectNotFound error is raised.
     """
 
-    object_designator: ObjectDesignatorDescription.Object
+    technique: DetectionTechnique
+    """
+    The technique that should be used for detection
+    """
+    state: DetectionState
```
tomsch420, line 472: this looks like the ORM also hasto be touched

src/pycrap/objects.py
```diff
@@ -82,7 +82,14 @@ class Cereal(Food):
...
+
 class Floor(PhysicalObject):
     """
     The lower surface of a room.
-    """
\ No newline at end of file
+    """
+
+
+class GenObj(PhysicalObject):
```
tomsch420, line 92: no abbrivations in class/method names

tomsch420 (changes requested):

PR #240, opened 2024-12-08
[CostmapLocation] for reachability should restrict to hand_links corr…

sunava:
…esponding to the selected arm.


src/pycram/designators/location_designator.py
```diff
@@ -186,7 +186,9 @@ def __iter__(self):
                 if self.reachable_for:
                     hand_links = []
                     for description in RobotDescription.current_robot_description.get_manipulator_chains():
-                        hand_links += description.end_effector.links
+                        if isinstance(self.reachable_arm, Arms):
```
Tigul, line 189: In this case hand_links would only be set if there is a reachable_arm set. Add an else branch for the case in which no reachable_arm ist set. 
If there is a reachable_arm set there is no need to iterate over all manupulator chains, so you can move the for loop in the else branch

Tigul (changes requested):

Tigul (approved):

PR #241, opened 2024-12-09
Icub dev

omareldardear:
Dear all, 

I create here the main pull request for the major updates of the integration of the iCub robot into pycram. 
In the next 2 days, I will replace some prints with exception handling and add some documentation. 

I create this pull request to initiate the process to speed up the integration. 
Please have a look and let me know if there are any reviews 

There are some requirements to have the code running. I will add all the details in README soon. 

Thanks


src/pycram/process_module.py
```diff
@@ -419,3 +419,9 @@ def close(self) -> Type[ProcessModule]:
         """
         raise NotImplementedError(
             f"There are no Process Modules for '{inspect.currentframe().f_code.co_name}' for robot '{self.robot_name}'")
+
+    def exit(self)->None:
```
Tigul, line 423: The exit method for Process modules is a nice idea. 
But the actual code to exit the connection to a robot should be directly in the Process module.
Please also add an exit method to the ProcessModule parent class and then call the currently loaded ProcessModule.exit method from this method

src/pycram/process_modules/icub_process_modules.py
```diff
@@ -0,0 +1,687 @@
...
+    DetectingMotion, MoveTCPMotion, MoveArmJointsMotion, WorldStateDetectingMotion, MoveJointsMotion, \
+    MoveGripperMotion, OpeningMotion, ClosingMotion
+from ..robot_description import RobotDescription
+from ..datastructures.world import World
+from ..world_concepts.world_object import Object
+from ..datastructures.pose import Pose
+from ..datastructures.enums import JointType, ObjectType, Arms, ExecutionType, GripperState
+from ..external_interfaces import giskard
+from ..external_interfaces.robokudo import *
+import yarp
```
Tigul, line 26: This add yarp as a hard dependency. Wrap this into a try, catch and add a log function in case the imports fail. Look [here](https://github.com/cram2/pycram/blob/dev/src/pycram/external_interfaces/giskard.py#L22) for reference

src/pycram/process_modules/icub_process_modules.py
```diff
@@ -0,0 +1,687 @@
...
+    yarp_bottle_msg.addVocab32('t', 'i', 'm', 'e')
+    yarp_bottle_msg.addFloat32(1.5)
+    yarp_bottle_msg.addVocab32('o', 'f', 'f')
+    yarp_bottle_msg.addInt32(7)
+    yarp_bottle_msg.addVocab32('p', 'o', 's')
+    target_loc: yarp.Bottle = yarp_bottle_msg.addList()
+    for i in hand_values:
+        target_loc.addFloat32(i)
+
+    print("command Ready to send to iCub  part tcp")
```
Tigul, line 44: Use pycram.ros.logging instead of print statements, throughout the whole file

src/pycram/process_modules/icub_process_modules.py
```diff
@@ -0,0 +1,687 @@
...
+            return False
+        elif reply_vocab == ACK_VOCAB:
+            print("ACK")
+            return True
+        else:
+            print("another reply")
+            return False
+
+HAND_CLOSED = [60.0,60.0,0,85.0,20,85,20,85,85]
+HAND_OPENED = [0   ,0   ,0,0   ,0 ,0 ,0 ,0 ,0]
```
Tigul, line 98: I guess these are joint values for the hand, they are already in the robot_description use these insetad and remove the global variable

src/pycram/process_modules/icub_process_modules.py
```diff
@@ -0,0 +1,687 @@
...
+from ..robot_description import RobotDescription
+from ..datastructures.world import World
+from ..world_concepts.world_object import Object
+from ..datastructures.pose import Pose
+from ..datastructures.enums import JointType, ObjectType, Arms, ExecutionType, GripperState
+from ..external_interfaces import giskard
+from ..external_interfaces.robokudo import *
+import yarp
+
+from pycram.yarp_utils.yarp_networking import NO_ACK_VOCAB, ACK_VOCAB, open_rpc_client_port, open_buffered_bottle_port, \
```
Tigul, line 28: I think it is better to move the calls to yarp into its own file and put this in external_interfaces instead of yarp_utils. So people know what purpose yarp has.

src/pycram/process_modules/icub_process_modules.py
```diff
@@ -0,0 +1,687 @@
...
+
+
+
+
+
+
+
+    def navigate(self):
+        if ProcessModuleManager.execution_type == ExecutionType.SIMULATED:
+            print('Navigate iCub')
```
Tigul, line 511: remove print

src/pycram/process_modules/icub_process_modules.py
```diff
@@ -0,0 +1,687 @@
...
+                                     [self.ctp_right_arm_client_port,self.ctp_left_arm_client_port])
+
+    def open(self):
+        print("iCub doesn't perform open action from here")
+
+    def close(self):
+        print("iCub doesn't perform close action from here")
+
+
+    def config_yarp_ports(self)->bool:
```
Tigul, line 572: the implementation of config methods should also go into an interface file to yarp instead of the process module manager and the just called from here

src/pycram/robot_description.py
```diff
@@ -167,6 +173,37 @@ def get_actuator_for_joint(self, joint: str) -> Optional[str]:
             return self.joint_actuators.get(joint)
         return None
 
+
+    def get_actuated_joint_names(self, part_name) -> Optional[str]:
+        """
+        Given a part name, returns the actuated joints.
+
+        Args:
```
Tigul, line 181: Stick to the rst format for documentation of classes. Look at the other methods for reference

src/pycram/robot_description.py
```diff
@@ -167,6 +173,37 @@ def get_actuator_for_joint(self, joint: str) -> Optional[str]:
             return self.joint_actuators.get(joint)
         return None
 
+
+    def get_actuated_joint_names(self, part_name) -> Optional[str]:
```
Tigul, line 177: Add typing for arguments

src/pycram/robot_description.py
```diff
@@ -167,6 +173,37 @@ def get_actuator_for_joint(self, joint: str) -> Optional[str]:
...
+        Args:
+            part_name (str): part name
+
+        Returns:
+            List of actuated joint names: or None
+        """
+
+        return self.joint_actuators.get(part_name)
+
+    def get_actuated_joint_indices(self, joint_name) -> Tuple[Optional[str],Optional[int], Optional[int]]:
```
Tigul, line 190: typing

src/pycram/robot_description.py
```diff
@@ -167,6 +173,37 @@ def get_actuator_for_joint(self, joint: str) -> Optional[str]:
...
+        """
+        Given a joint name, returns the indices of the part and the joint within that part.
```
Tigul, line 192: I don't get what this method does. What is a "part"?

src/pycram/yarp_utils/yarp_joint_state_updater.py
```diff
@@ -0,0 +1,324 @@
+import math
+import sys
+from time import sleep
+
+from pycram import World
```
Tigul, line 5: Change this import to "from ..datastructures.world import World"

src/pycram/yarp_utils/yarp_joint_state_updater.py
```diff
@@ -0,0 +1,324 @@
...
+from pycram.robot_description import RobotDescription
+from pycram.yarp_utils.yarp_networking import interrupt_and_close
```
Tigul, line 7: Change these to relative imports

src/pycram/yarp_utils/yarp_joint_state_updater.py
```diff
@@ -0,0 +1,324 @@
...
+import yarp
+
+
+class icub_state_updater(yarp.RFModule):
```
Tigul, line 11: For the whole class: Add docstrings, typehints for arguments and return 

The python convention for class names is CamelCase, please rename the class

src/pycram/yarp_utils/yarp_joint_state_updater.py
```diff
@@ -0,0 +1,324 @@
...
+
+        self.state_torso_port_name = self.getName("/state/torso:i")
+        self.state_right_arm_port_name = self.getName("/state/right_arm:i")
+        self.state_left_arm_port_name = self.getName("/state/left_arm:i")
+        self.state_head_port_name = self.getName("/state/head:i")
+
+
+        # Create handle port to read message
+        if not self.handle_port.open('/' + self.module_name):
+            print("Can't open the port correctly")
```
Tigul, line 56: Use pycram.ros.logging instead of prints, also all prints below. I really like the precise checking if all ports can be  opened and the corresponding error logging

src/pycram/yarp_utils/yarp_joint_state_updater.py
```diff
@@ -0,0 +1,324 @@
...
+            return False
+
+
+        self.torso_joints_names =  RobotDescription.current_robot_description.get_actuated_joint_names("torso")
+        self.right_arm_joints_names = RobotDescription.current_robot_description.get_actuated_joint_names("right_arm")
+        self.left_arm_joints_names = RobotDescription.current_robot_description.get_actuated_joint_names("left_arm")
+        self.head_joints_names = ["neck_pitch","neck_roll","neck_yaw","eyes_tilt","r_eye_pan_joint","l_eye_pan_joint"]
+
+        if self.torso_joints_names is None:
+            print("Error in joint names")
```
Tigul, line 85: That is a bit misleading, just say there are no joint names defined in the description

src/pycram/yarp_utils/yarp_joint_state_updater.py
```diff
@@ -0,0 +1,324 @@
...
+        if command.get(0).asString() == "quit":
+            reply.addString("quitting")
+            return False
+        else:
+            reply.addString("other request")
+
+        return True
+
+
+    def getPeriod(self):
```
Tigul, line 166: The convention for python methods is snake_case, please rename to "get_period"

src/pycram/yarp_utils/yarp_joint_state_updater.py
```diff
@@ -0,0 +1,324 @@
...
+                                       arm_bottle.get(15).asFloat64()/3.0)
+        self.update_joint_degree(f"{arm}_hand_little_3_joint",
+                                       arm_bottle.get(15).asFloat64()/3.0)
+        ####################################################################################
+
+
+
+
+
+    def updateModule(self):
```
Tigul, line 278: rename to "update_module"

src/pycram/yarp_utils/yarp_joint_state_updater.py
Tigul, line 1: As already mentioned, move the calls directly to yarp and the configuartin of the network into its own file in pycram.external_interfaces.yarp

src/pycram/yarp_utils/yarp_networking.py
Tigul, line 1: Also put these methods in pycram.external_interfaces.yarp

src/pycram/yarp_utils/yarp_networking.py
```diff
@@ -0,0 +1,36 @@
+
+import yarp
+
+
+ACK_VOCAB = yarp.createVocab32('a','c','k')
+NO_ACK_VOCAB = yarp.createVocab32('n','a','c','k')
+
+
+def init_yarp_network():
```
Tigul, line 9: Also here, please add doc strings and type hints

test/test_icub.py
```diff
@@ -0,0 +1,116 @@
...
+        # cls.milk = Object("milk", pycrap.Milk, "milk.stl", pose=Pose([1.3, 1, 0.9]))
+        cls.robot = Object("icub", pycrap.Robot, "icub" + cls.extension, pose=Pose([0, 0, 0.5]))
+        cls.kitchen = Object("kitchen", pycrap.Kitchen, "kitchen" + cls.extension)
+        # cls.cereal = Object("cereal", pycrap.Cereal, "breakfast_cereal.stl",
+        #                     pose=Pose([1.3, 0.7, 0.95]))
+        ProcessModule.execution_delay = False
+        ProcessModuleManager.execution_type = ExecutionType.SIMULATED
+        cls.viz_marker_publisher = VizMarkerPublisher()
+
+    def test_move_joint_direct(self):
```
Tigul, line 50: There is no assert here

test/test_icub.py
```diff
@@ -0,0 +1,116 @@
...
+        self.robot.set_joint_position("l_hand_thumb_0_joint",math.radians(30))
+
+
+    def test_move_joints_simulated(self):
+        with simulated_robot:
+            MoveJointsMotion(["torso_roll"], [math.radians(-20.0)]).perform()
+            sleep(3)
```
Tigul, line 57: Are these sleeps necessary? Are the calls in yarp asynchronous?

test/test_icub.py
```diff
@@ -0,0 +1,116 @@
...
+            LookingMotion(Pose([-2, -2, 3])).perform()
+            sleep(3)
+            MoveGripperMotion(GripperState.OPEN, Arms.RIGHT).perform()
+            MoveGripperMotion(GripperState.OPEN, Arms.LEFT).perform()
+            sleep(20)
+            MoveGripperMotion(GripperState.CLOSE, Arms.RIGHT).perform()
+            MoveGripperMotion(GripperState.CLOSE, Arms.LEFT).perform()
+            sleep(10)
+            MoveGripperMotion(GripperState.OPEN, Arms.RIGHT).perform()
+            MoveGripperMotion(GripperState.OPEN, Arms.LEFT).perform()
```
Tigul, line 73: Also no asserts here. So you are only executing the motions and checking if nothing crashes

test/test_icub.py
```diff
@@ -0,0 +1,116 @@
...
+                interrupt_and_close(yarpModule)
+
+
+    def test_park_arms(self):
+        with real_robot:
+            ParkArmsActionPerformable(arm=Arms.BOTH).perform()
+
+
+    def test_try_world(self):
+        sleep(100)
```
Tigul, line 114: What? Why?
I guess this was for testing, in this case please delete it

test/test_icub.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+
+
+
+    def test_move_joints(self):
+
+
+        with real_robot:
+            yarpModule = run_icub_state_updater(['/', '--robot', 'icubSim'])
```
Tigul, line 83: There is unittest.skip for this case, look [here](https://github.com/cram2/pycram/blob/dev/test/test_multiverse.py#L34) for reference

Tigul (changes requested):

src/pycram/process_module.py
omareldardear: isn't this what i did ? 
Would you explain more this part

src/pycram/process_modules/icub_process_modules.py
omareldardear: not all the joint are directly actuated 1:1. 
In the robot description I have the values for every joint for close and open. However the control joints are different and also different values. 
So this list is the actuated control values. which is only used for the real robot 
What do you think ?

src/pycram/process_modules/icub_process_modules.py
omareldardear: Done

src/pycram/process_modules/icub_process_modules.py
omareldardear: Done

src/pycram/process_modules/icub_process_modules.py
omareldardear: Done

test/test_icub.py
omareldardear: I deleted the whole file, this was just for me to test the implementation. 
I think I will try to make jupyter notebook on virtual labs for this instead of here

test/test_icub.py
omareldardear: no, not necessary, it was just too fast to see it in the simulation. Anyways, the whole file os removed :)

test/test_icub.py
omareldardear: same

test/test_icub.py
omareldardear: same

test/test_icub.py
omareldardear: same

src/pycram/robot_description.py
omareldardear: done

src/pycram/robot_description.py
omareldardear: ?

src/pycram/robot_description.py
omareldardear: There is a dict of part_name , actuated joints list. 
Part can be (torso, left_arm,right_arm,head) in the icub case. a part can have multiple kinematic chains. 
for example because of the fingers or binocular vision. 

For icub, each part is controlled by a single command.

src/pycram/yarp_utils/yarp_joint_state_updater.py
omareldardear: done

src/pycram/yarp_utils/yarp_joint_state_updater.py
omareldardear: done

src/pycram/yarp_utils/yarp_joint_state_updater.py
omareldardear: done

src/pycram/yarp_utils/yarp_joint_state_updater.py
omareldardear: done

src/pycram/yarp_utils/yarp_joint_state_updater.py
omareldardear: I had it for some tests, but no need for it. you are right

src/pycram/yarp_utils/yarp_joint_state_updater.py
omareldardear: It is a defined method in the parent class. I can't change it. 
It is also written in this way because yarp is written in C++ with a binding in pythin.

src/pycram/yarp_utils/yarp_networking.py
omareldardear: done

src/pycram/yarp_utils/yarp_networking.py
omareldardear: done

src/pycram/process_modules/icub_process_modules.py
sunava, line 1: Please checkout the new default_process_modules and how they are handled

src/pycram/process_modules/icub_process_modules.py
omareldardear: In this configuration function, I open and connect specific ports to control and read the status of the robot. It doesn't configure the network or general yarp things. It is more related to the direct low-level control of the robot. 

IIf you still think that it should be somehow in the external_interface, I can try to restructure the code for that. 
But it is kinda a major change. Because it will change the general pipeline of the code and its initialization

src/pycram/process_modules/icub_process_modules.py
omareldardear: I didn't change the default processing modules for the simulated robot. 
I used them as it is. 

For the real robot, I implemented the specific functionalities to control icub in the right way. 
Note: the tcp/looking/ etc poses are expected to be w.r.t. robot frame. At the moment as we don't have a simulated world for the real robot. maybe lateron, we can easily do the transformation. 

If there is any other notes, please clarify what you mean. 
Thanks

src/pycram/process_module.py
Tigul: Yes what you did is correct, I was just confused about the structure of the ProcessModuleManager

src/pycram/process_modules/icub_process_modules.py
Tigul: Ah ok, then you cloud put them in the move_gripper process module of the real robot. This way it should be clear where they are used.

src/pycram/robot_description.py
Tigul: Sorry,  I meant add typing to the parameter of the function

src/pycram/robot_description.py
Tigul: I see, could you add that to the doc string?

src/pycram/external_interfaces/yarp_networking.py
```diff
@@ -0,0 +1,66 @@
...
+
+ACK_VOCAB = yarp.createVocab32('a','c','k')
+NO_ACK_VOCAB = yarp.createVocab32('n','a','c','k')
+
+
+def init_yarp_network():
+    """
+        Initializes the YARP network.
+
+        **Returns:**
```
Tigul, line 14: Please stick to the rst standard

src/pycram/external_interfaces/yarp_networking.py
```diff
@@ -0,0 +1,66 @@
+from typing_extensions import Tuple,Optional
+
+import yarp
```
Tigul, line 3: I think this needs a try, catch block

src/pycram/process_modules/icub_process_modules.py
```diff
@@ -0,0 +1,686 @@
...
+        loginfo("Received: NO_ACK")
+        return False
+    elif reply_vocab == ACK_VOCAB:
+        loginfo("Received: ACK")
+        return True
+    else:
+        loginfo("Received: another reply")
+        return False
+
+def update_part(state_port,ctp_port, joint_to_change_idx, joints_to_change_pos):
```
Tigul, line 50: Please add documentation here

src/pycram/process_modules/icub_process_modules.py
```diff
@@ -0,0 +1,686 @@
...
+        loginfo("iCubNavigationReal function. No implementation ATM")
+
+
+class iCubMoveHeadReal(ProcessModule):
+    """
+        This process module moves the head to look at a specific point in the world coordinate frame.
+        This point can either be a position or an object.
+        """
+
+    def __init__(self,lock,cmd_port:yarp.RpcClient):
```
Tigul, line 102: The type annotation yarp.RpcClient will not work if yarp is not there. You can put it in quotation marks to prevent the interpreter evaluating it

omareldardear:
I noticed that the CI/CD checks failed because of the imports of yarp. 
I did a fix and pushed the changes.

src/pycram/external_interfaces/yarp_networking.py
omareldardear: Done

src/pycram/process_modules/icub_process_modules.py
omareldardear: done

omareldardear:
I went through the code again and all your comments. 
You can check again now

PR #252, opened 2024-12-16
Added fitting floor for apartment

pkehr:
Spawning this floor can be toggled by adjusting the boolean parameter "spawn_floor" in the VizMarkerPublisher


Tigul:
Why is this PR necessary? The floor will always be spawned at the start of each world.

PR #253, opened 2025-01-14
Find non-fixed parent joint for link.

artnie:
This removes the joint type when searching for the movable parent joint of a link, e.g. a handle, finding the closest non-fixed joint.

I also adjusted the tests and all occurrences.

@AbdelrhmanBassiouny could you have a look?


tomsch420:
Gucci

PR #254, opened 2025-01-14
Fix joint for fridge door

artnie:
Change the door joints limits and from prismatic to revolute, similar to https://github.com/code-iai/iai_maps/pull/118

@hawkina since I asked you to look into the iai_maps PR already, could you verify this one too please?


tomsch420:
Ah yes, that is correct. Gucci!

PR #255, opened 2025-01-15
Add grasps BOTTOM and BACK

artnie:
Adds the BACK and BOTTOM grasp to the dataclass and PR2 description. Necessary for picking objects out of the fridge that are spawned with default rotation.

Also uses the normalized quaternions.


tomsch420:
Fine

PR #256, opened 2025-01-15
Semantic representation

sunava:
The two functions are essentially designed to make the action more semantic, as requested and discussed.


tomsch420:
I like this. I think micheal would also like this.

PR #257, opened 2025-01-16
Arm fixes

Tigul:
Fixed the handling of ParkArms and MoveGripper of the input Arms.BOTH and fixes issues when spawning a mesh object with the parameter scale_mesh and color


tomsch420:
gucci

PR #258, opened 2025-01-17
Merge bounding boxes and Merge Descriptions

AbdelrhmanBassiouny:
If we have multible bouding boxes lets say for multi link objects and want to get a bounding box for a combination or all of this links, this is a way of combining them into a single mesh object.

Currently, it is done in two ways, one is using RandomEvents, and the other is just using all the faces of all the bounding boxes together and making a mesh out of it.

Merging descriptions is basically merging two objects into one object by merging their urdf files into one and attaching them to each other, this is useful when you want to think of two attached objects as one object instead of two.

### DIFF
--- a/src/pycram/datastructures/world.py
+++ b/src/pycram/datastructures/world.py
@@ -412,7 +412,7 @@ class World(WorldEntity, ABC):
         """
         pass
 
-    def remove_object(self, obj: Object, remove_from_simulator: bool = True) -> None:
+    def remove_object(self, obj: Object) -> None:
         """
         Remove this object from the current world.
         For the object to be removed it has to be detached from all objects it
@@ -425,7 +425,7 @@ class World(WorldEntity, ABC):
 
         obj.detach_all()
 
-        if remove_from_simulator and self.remove_object_from_simulator(obj):
+        if self.remove_object_from_simulator(obj):
             self.objects.remove(obj)
             self.remove_object_from_original_state(obj)
 
--- a/src/pycram/datastructures/world.py
+++ b/src/pycram/datastructures/world.py
@@ -412,7 +412,7 @@ class World(WorldEntity, ABC):
         """
         pass
 
-    def remove_object(self, obj: Object) -> None:
+    def remove_object(self, obj: Object, remove_from_simulator: bool = True) -> None:
         """
         Remove this object from the current world.
         For the object to be removed it has to be detached from all objects it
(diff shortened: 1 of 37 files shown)


src/pycram/datastructures/world.py
```diff
@@ -412,7 +412,7 @@ def remove_object_from_simulator(self, obj: Object) -> bool:
         """
         pass
 
-    def remove_object(self, obj: Object) -> None:
+    def remove_object(self, obj: Object, remove_from_simulator: bool = True) -> None:
```
Tigul, line 415: Please add the new parameter to the doc string

src/pycram/datastructures/world.py
```diff
@@ -425,7 +425,7 @@ def remove_object(self, obj: Object) -> None:
 
         obj.detach_all()
 
-        if self.remove_object_from_simulator(obj):
+        if remove_from_simulator and self.remove_object_from_simulator(obj):
```
Tigul, line 428: I'm a little confused about the remove_from_simulator arg, if it is set to False the method does nothing except detaching all object from this object

src/pycram/world_concepts/world_object.py
```diff
@@ -387,6 +382,7 @@ def _spawn_object_and_get_id(self) -> int:
             logging.error(
                 "The File could not be loaded. Please note that the path has to be either a URDF, stl or obj file or"
                 " the name of an URDF string on the parameter server.")
+            logging.error(e)
```
Tigul, line 385: Could you change both logging statements to the ros.logging statements

src/pycram/datastructures/world.py
AbdelrhmanBassiouny: removed the parameter altogether :D

src/pycram/datastructures/world.py
AbdelrhmanBassiouny: so I removed the parameter, but the thing is in Multiverse, environments cannot be removed (or not recommended to do so) and thus they will not be removed, so in this case the `self.remove_object_from_simulator(obj)` will return False, and to let pycram have knowledge of this, I keep the object in the world list of objects (of pycram) and I do not remove it from the original state.

src/pycram/world_concepts/world_object.py
AbdelrhmanBassiouny: Done

src/pycram/datastructures/world.py
Tigul: Ah ok, in that case the parameter makes sense

PR #259, opened 2025-01-17
Add gripper types to various robot descriptions

J-Schaefer:
Modified descriptions of Tiago, Boxy, UR5, HSR,
Donbot, Stretch.

### DIFF
--- a/src/pycram/robot_descriptions/boxy_description.py
+++ b/src/pycram/robot_descriptions/boxy_description.py
@@ -1,7 +1,7 @@
 from ..ros.ros_tools import  get_ros_package_path
 from ..robot_description import RobotDescription, CameraDescription, KinematicChainDescription, \
     EndEffectorDescription, RobotDescriptionManager
-from ..datastructures.enums import Arms, Grasp, GripperState, TorsoState
+from ..datastructures.enums import Arms, Grasp, GripperState, TorsoState, GripperType
 
 filename = get_ros_package_path('pycram') + '/resources/robots/' + "boxy" + '.urdf'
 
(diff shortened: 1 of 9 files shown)


src/pycram/robot_descriptions/boxy_description.py
```diff
@@ -29,7 +29,8 @@
 
 right_gripper.add_static_joint_states(GripperState.OPEN, {"right_gripper_joint": 0.548})
 right_gripper.add_static_joint_states(GripperState.CLOSE, {"right_gripper_joint": 0.0})
-
+right_gripper.end_effector_type = GripperType.PARALLEL  # current gripper in sim, change later
+right_gripper.opening_distance = 0.548
```
Tigul, line 33: I don't think this value is correct. That would mean the distance between the grippers is over half a meter. 
The values given in the static joint chains are meant for the joint itself, which in this case would be radiants for the joint.

src/pycram/robot_descriptions/boxy_description.py
```diff
@@ -53,7 +54,8 @@
 
 left_gripper.add_static_joint_states(GripperState.OPEN, {"left_gripper_joint": 0.548})
 left_gripper.add_static_joint_states(GripperState.CLOSE, {"left_gripper_joint": 0.0})
-
+left_gripper.end_effector_type = GripperType.PARALLEL  # current gripper in sim, change later
+left_gripper.opening_distance = 0.548
```
Tigul, line 58: See above commentat

src/pycram/robot_descriptions/donbot_description.py
```diff
@@ -27,7 +27,8 @@
 
 right_gripper.add_static_joint_states(GripperState.OPEN, {'gripper_joint': 0.0})
 right_gripper.add_static_joint_states(GripperState.CLOSE, {'gripper_joint': 0.2})
-
+right_gripper.end_effector_type = GripperType.PARALLEL
+right_gripper.opening_distance = 0.2
```
Tigul, line 31: Same here

src/pycram/robot_descriptions/stretch_description.py
```diff
@@ -31,7 +31,8 @@
                                                                 'joint_gripper_finger_right': 0.59})
 gripper_description.add_static_joint_states(GripperState.CLOSE, {'joint_gripper_finger_left': 0.0,
                                                                  'joint_gripper_finger_right': 0.0})
-
+gripper_description.end_effector_type = GripperType.PARALLEL
+gripper_description.opening_distance = 0.59
```
Tigul, line 35: Also to big

J-Schaefer:
@Tigul Fixed all values now. Also for PR2. Looked up the real values or measured.

PR #261, opened 2025-01-22
Smaller improvements

Tigul:
# Description 
Adds some minor improvements:
 * Added load and unload methods to the RobotDescription which allow to execute code upon loading of the respective descriptin 
 * Added support for the Kevin support (description and process modules)
 * Added a function and button in the BulletWorld GUI to save the current camera config as default for the future.
 * Added decorator for the pre and post perform methods, so functions can be annotated and then be executed before or after an action


config/world_conf.py
```diff
@@ -109,6 +111,18 @@ class WorldConfig:
     Whether the depth images produced by :meth:`datastructures.world.World.get_images_for_target` are in meters.
     """
 
+    @classproperty
+    def default_camera_config(self) -> Dict:
```
tomsch420, line 115: cls

src/pycram/designator.py
```diff
@@ -190,9 +201,11 @@ def perform(self) -> Any:
 
             :return: The result of the action in the plan
             """
-            self.pre_perform()
+            for pre_perform in ActionDesignatorDescription.Action._pre_perform_callbacks:
+                pre_perform(self)
             result = self.plan()
```
tomsch420, line 206: wrap this with try except PlanError

src/pycram/process_modules/kevin_process_modules.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+class KevinNavigate(DefaultNavigation):
+    """
+    Process module for the simulated Kevin that sends a cartesian goal to the robot to move the robot base
+    """
+    pass
+
+
+class KevinMoveHead(ProcessModule):
```
tomsch420, line 21: this is not okidoki

src/pycram/utils.py
```diff
@@ -535,4 +535,38 @@ def map_color_names_to_rgba(name: str) -> Color:
         "black": Color(0, 0, 0, 1),
         "grey": Color(0.5, 0.5, 0.5, 1),
     }
+
     return colors.get(name.lower(), Color(0, 0, 0, 1)).to_list()  # Fallback to black
+
+
+
+class ClassPropertyDescriptor(object):
```
tomsch420, line 543: inheritance from object is unnescessary

src/pycram/utils.py
```diff
@@ -535,4 +535,38 @@ def map_color_names_to_rgba(name: str) -> Color:
...
+        return self.fset.__get__(obj, type_)(value)
+
+    def setter(self, func):
+        if not isinstance(func, (classmethod, staticmethod)):
+            func = classmethod(func)
+        self.fset = func
+        return self
+
+def classproperty(func):
+    if not isinstance(func, (classmethod, staticmethod)):
```
tomsch420, line 567: i would not do that here, this may hide a missuse of  class decorator

tomsch420 (changes requested):

src/pycram/designator.py
```diff
@@ -203,9 +204,13 @@ def perform(self) -> Any:
...
                 pre_perform(self)
-            result = self.plan()
-            for post_perform in ActionDesignatorDescription.Action._post_perform_callbacks:
-                post_perform(self)
+            try:
+                result = self.plan()
+            except PlanFailure as e:
+                raise e
+            finally:
+                for post_perform in ActionDesignatorDescription.Action._post_perform_callbacks:
```
tomsch420, line 212: cant you use self.post_perform_callbacks here?

Tigul: done

tomsch420 (approved):

PR #264, opened 2025-01-30
Manipiulator description

AbdelrhmanBassiouny:
depends on #258 

Created a ManipulatorData dataclass to clean and simplify creating one arm manipulator descriptions in the robot_descriptions folder.

Added methods that use the Manipulator data to create the robot description with the arm and the gripper.

### DIFF
--- a/src/pycram/datastructures/dataclasses.py
+++ b/src/pycram/datastructures/dataclasses.py
@@ -945,15 +945,6 @@ class VirtualMobileBaseJoints:
 
 
 @dataclass
-class MultiverseBoundingBox:
-    """
-    Dataclass for storing the bounding box of a body in the Multiverse simulation.
-    """
-    min_point: List[float]
-    max_point: List[float]
-
-
-@dataclass
 class MultiverseMetaData:
     """Meta data for the Multiverse Client, the simulation_name should be non-empty and unique for each simulation"""
     world_name: str = "world"
@@ -980,15 +971,6 @@ class RayResult:
         return: Whether the ray intersects with a body.
         """
         return self.distance >= 0 and self.body_name != ""
-
-
-@dataclass
-class MultiverseObjectContactData:
-    """
-    A dataclass to store all the contact data returned from Multiverse for a single object.
-    """
-    body_names: List[str]
-    data: List[MultiverseContactPoint]
 
 
 @dataclass
--- a/src/pycram/datastructures/dataclasses.py
+++ b/src/pycram/datastructures/dataclasses.py
@@ -710,6 +710,10 @@ class ContactPoint:
     lateral_friction_1: Optional[LateralFriction] = None
     lateral_friction_2: Optional[LateralFriction] = None
 
+    @property
+    def normal(self) -> List[float]:
+        return self.normal_on_body_b
+
     def __str__(self):
         return f"ContactPoint: {self.body_a.name} - {self.body_b.name}"
 
@@ -795,7 +799,7 @@ class ContactPointsList(list):
         :param obj: An instance of the Object class that represents the object.
         :return: A list of Link instances that represent the links in contact of the object.
         """
-        return [point.body_b for point in self if point.body_b.parent == obj]
+        return [point.body_b for point in self if point.body_b.parent_entity == obj]
 
     def get_points_of_object(self, obj: Object) -> 'ContactPointsList':
         """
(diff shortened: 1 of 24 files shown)


src/pycram/datastructures/dataclasses.py
```diff
@@ -25,6 +25,31 @@
     from .world_entity import PhysicalBody
 
 
+@dataclass
+class ManipulatorData:
```
Tigul, line 29: Could you add some short doc to the parameters, that would make it easier to define them

src/pycram/robot_description.py
```diff
@@ -161,6 +162,28 @@ def __init__(self, name: str, base_link: str, torso_link: str, torso_joint: str,
...
+                arm_home_values: Optional[Dict[str, float]] = None) -> KinematicChainDescription:
+        """
+        Creates and adds an arm to the RobotDescription.
+
+        :param end_link: Last link of the arm
+        :param arm_type: Type of the arm
+        :param arm_name: Name of the arm
+        :param arm_home_values: Dictionary of joint names and their home values
+        """
+        arm_start = self.base_link if self.torso_link == '' else self.torso_link
```
Tigul, line 177: I don't think this is a good idea, arms might not be attached to the torso link directly and defining the whole robot as an arm chain could have consequences that are hard to debug.

src/pycram/robot_description.py
```diff
@@ -161,6 +162,28 @@ def __init__(self, name: str, base_link: str, torso_link: str, torso_joint: str,
...
+                arm_type: Arms = Arms.RIGHT,
+                arm_name: str = "manipulator",
+                arm_home_values: Optional[Dict[str, float]] = None) -> KinematicChainDescription:
+        """
+        Creates and adds an arm to the RobotDescription.
+
+        :param end_link: Last link of the arm
+        :param arm_type: Type of the arm
+        :param arm_name: Name of the arm
+        :param arm_home_values: Dictionary of joint names and their home values
```
Tigul, line 175: What are home values?

src/pycram/robot_description.py
```diff
@@ -469,6 +492,45 @@ def _init_joints(self):
         self.joint_names = list(filter(lambda j: self.urdf_object.joint_map[j].type != JointType.FIXED
                                                  or self.include_fixed_joints, joints))
 
+    def create_end_effector(self,
```
Tigul, line 495: Isn't this just a wrapper around the constructor of an EndEffectorDescription?

src/pycram/robot_description.py
```diff
@@ -721,3 +782,29 @@ def joints(self) -> List[str]:
         :return: List of joint names
         """
         return self.joint_names
+
+
+def create_manipulator_description(data: ManipulatorData,
```
Tigul, line 787: We should stick to either manipulator or end effector, having both interchangeably could be confusing

src/pycram/robot_descriptions/panda_description.py
```diff
@@ -0,0 +1,41 @@
...
+    gripper_relative_dir=None,
+    gripper_tool_frame="right_pad",
+
+    gripper_joint_names=[f'finger_joint{i}' for i in [1, 2]],
+    closed_joint_values=[0.0, 0.0],
+    open_joint_values=[0.04, 0.04],
+
+    gripper_cmd_topic="/gripper_command",
+    gripper_open_cmd_value=0.0,
+    gripper_close_cmd_value=255.0)
```
Tigul, line 25: Are you sure about this value?

Tigul (changes requested):
Also check the failing CI

src/pycram/robot_description.py
AbdelrhmanBassiouny: ok I see, I now added arm_start as an optional argument.

src/pycram/robot_description.py
AbdelrhmanBassiouny: The default configuration or home position of the arm (e.g. park arms), add more clarification to its doc now.

src/pycram/robot_description.py
AbdelrhmanBassiouny: well manipulator is arm plus end_effector

src/pycram/robot_descriptions/panda_description.py
AbdelrhmanBassiouny: for the ur5 yes, for other robots no, XD

AbdelrhmanBassiouny: I can None them for other robots

src/pycram/robot_description.py
Tigul: ok that makes sense

Tigul (approved):

PR #265, opened 2025-02-06
Non convex goals

tomsch420:
Added a way to create paths that go through non-convex spaces. 
This is done by calculating the complement of all obstacles and writing down what can be reached from where in the ConnectivityGraph datastructure. To find a path from point to point, the shortest path (if it exists) is used in the connectivity graph.


AbdelrhmanBassiouny (approved):
Gucci

src/pycram/ros_utils/viz_marker_publisher.py
```diff
@@ -328,3 +329,61 @@ def clear_all_marker(self):
...
+class TrajectoryPublisher:
+    """
+    Publishes a trajectory as a MarkerArray to visualize it in rviz.
+    """
+
+    @cached_property
+    def publisher(self):
+        pub = create_publisher("/pycram/trajectory", MarkerArray)
+
+        time.sleep(0.5)
```
Tigul, line 344: Maybe add a comment that this is necessary

src/pycram/graph_of_convex_sets.py
```diff
@@ -0,0 +1,448 @@
...
+                   z=intersection[self.z].simple_sets[0])
+
+    def __hash__(self):
+        return hash(id(self.simple_event))
+
+    @classmethod
+    def from_event(cls, event: Event) -> List[Self]:
+        result = []
+        for simple_event in event.simple_sets:
+            simple_event: SimpleEvent
```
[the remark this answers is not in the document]
tomsch420: it was just for me to see the linting :D event.simple_sets is per definition a sorted set of abstract simple sets. SimpleEvent is an implementation of that with extra features

src/pycram/datastructures/dataclasses.py
```diff
@@ -279,6 +407,33 @@ def get_max(self) -> List[float]:
         """
         return [self.max_x, self.max_y, self.max_z]
 
+    def enlarge(self, min_x: float = 0., min_y:float = 0, min_z: float = 0,
+                max_x: float = 0., max_y: float = 0., max_z: float = 0.):
+        """
+        Enlarge the axis-aligned bounding box by a given amount in-place.
```
Tigul, line 413: New line

Tigul (approved):

PR #266, opened 2025-02-11
Ray test datastructure

AbdelrhmanBassiouny:
[RayTest] Added a data structure for result of ray_test and ray_test_batch plus an option for ray distance calculation.


src/pycram/datastructures/dataclasses.py
```diff
@@ -1080,6 +1080,71 @@ class MultiverseMetaData:
 
 @dataclass
 class RayResult:
+    """
+    A dataclass to store the ray result. The ray result contains the body name that the ray intersects with and the
+    distance from the ray origin to the intersection point.
+    """
```
tomsch420, line 1086: formatting (press ctrl + alt + shift + l)

tomsch420 (changes requested):

tomsch420 (approved):

PR #267, opened 2025-02-12
Ros2

Tigul:
# Description 
Adapts PyCRAM for ROS2 and Ubuntu24.04 this includes:
*  an abstraction package to abstract from ROS1/ROS2 differences. 
* A new ik solver which is directly integrated in PyCRAM 
* Changes to the way ROS messages are initialized 
* The LocalTransformer was updated to use TF2 instead of TF1
* Some changes to the imports since TF is not available 
* New structure and manifest 
* New setup.py
* New way of discovering other ROS packages


Tigul:
Closing this due to complicated merge

PR #268, opened 2025-02-13
[PyCrap] define joint ontology in joint class, HasConcept is in WorldEntity

AbdelrhmanBassiouny:
[PyCrap] moved update of joint ontology individual to Joint class, moved HasConcept from PhysicalBody to WorldEntity so as to make joints also have a concept as they inherit from WorldEntity not from PhysicalBody, it also makes sense as every thing in the world should theoretically have a concept.


Tigul:
Looks good, I  pushed an update to dev that should fix the failing CI if you could merge the dev we can merge this later

tomsch420 (approved):

PR #269, opened 2025-02-13
Ros2 new

Tigul:
# Description

Adapts PyCRAM for ROS2 and Ubuntu24.04 this includes:

 *   an abstraction package to abstract from ROS1/ROS2 differences.
 *  A new ik solver which is directly integrated in PyCRAM
 * Changes to the way ROS messages are initialized
 *  The LocalTransformer was updated to use TF2 instead of TF1
 *  Some changes to the imports since TF is not available
 *  New structure and manifest
 *  New setup.py
 *  New way of discovering other ROS packages
 * Included tf_transformations since it is not released for ROS1


setup.py
```diff
@@ -1,12 +1,53 @@
...
+        packages=find_packages(exclude=['test']),
+        data_files=[
+            ('share/ament_index/resource_index/packages',
+                ['resource/' + package_name]),
+            ('share/' + package_name, ['package.xml']),
+        ] + generate_data_files('share/' + package_name + '/', 'resources'),
+        install_requires=['setuptools'],
+        zip_safe=True,
+        maintainer='jonas',
+        maintainer_email='j.dech@gmx.net',
```
tomsch420, line 45: shouldnt this be your uni mail?

src/pycram/external_interfaces/pinocchio_ik.py
```diff
@@ -0,0 +1,174 @@
...
+
+    JOINT_ID = model.frames[model.getFrameId(target_link)].parent
+    # Object to destination transformation
+    oMdes = pinocchio.XYZQUATToSE3(np.array(target_pose.position_as_list() + target_pose.orientation_as_list()))
+
+    # Initial joint configuration
+    # q = pinocchio.neutral(model)
+    q = create_joint_configuration(robot, model)
+    eps = 1e-4
+    IT_MAX = 10000
```
tomsch420, line 52: shouldnt these be parameters of a class or config file?

src/pycram/external_interfaces/pinocchio_ik.py
```diff
@@ -0,0 +1,174 @@
...
+    :param target_transformation: The target transformation.
+    :param eps: The error threshold.
+    :param max_iter: The maximum number of iterations.
+    :param dt: The time step.
+    :param damp: The damping factor.
+    :return: The final joint configuration.
+    """
+    q = configuration
+    i = 0
+    while True:
```
tomsch420, line 82: perhaps do this without while true? its always an eyebrow raiser

src/pycram/external_interfaces/pinocchio_ik.py
```diff
@@ -0,0 +1,174 @@
...
+        err = iMd.translation
+        if norm(err) < eps:
+            success = True
+            break
+        if it >= max_iter:
+            success = False
+            break
+        J = pinocchio.computeJointJacobian(model, data, q, target_joint_id)  # in joint frame
+        J = -J[:3, :]  # linear part of the Jacobian
+        v = -J.T.dot(solve(J.dot(J.T) + damp * np.eye(3), err))
```
tomsch420, line 133: abstract the error metric from the rest of the algorithm. Time for a class?

src/pycram/external_interfaces/pinocchio_ik.py
```diff
@@ -0,0 +1,174 @@
...
+    :param model: The Pinocchio model.
+    :param configuration: The initial joint configuration.
+    :param data: The Pinocchio data.
+    :param target_joint_id: The target joint ID.
+    :param target_transformation: The target transformation.
+    :param eps: The error threshold.
+    :param max_iter: The maximum number of iterations.
+    :param dt: The time step.
+    :param damp: The damping factor.
+    :return: The final joint configuration.
```
tomsch420, line 117: and a bool indicating success

src/pycram/external_interfaces/pinocchio_ik.py
```diff
@@ -0,0 +1,174 @@
...
```
tomsch420, line 78: and a bool indicating success

src/pycram/external_interfaces/pinocchio_ik.py
```diff
@@ -0,0 +1,174 @@
...
+
+def clip_joints_to_limits(joint_positions: List[float], joint_limits: List[Tuple[float, float]]) -> np.ndarray[float]:
+    """
+    Clip the joint positions to the joint limits.
+
+    :param joint_positions: The joint positions to clip.
+    :param joint_limits: The joint limits to clip to as a list of tuples (min, max).
+    :return: The clipped joint positions.
+    """
+    return np.array([min(max(joint_position, joint_limit[0]), joint_limit[1]) for joint_position, joint_limit in zip(joint_positions, joint_limits)])
```
tomsch420, line 174: numpy.clip exists

src/pycram/ros/ros2/__init__.py
```diff
@@ -0,0 +1,23 @@
+# Funky MonkeyPatching for ROS2 compatability
+import builtin_interfaces.msg
+def to_sec(self):
+    return self.sec
```
tomsch420, line 4: document this please :D

src/pycram/ros/ros2/data_types.py
```diff
@@ -0,0 +1,28 @@
+import builtin_interfaces.msg
+import rclpy
+from . import node
+import builtin_interfaces
+
+class Time(builtin_interfaces.msg.Time):
```
tomsch420, line 6: doc

src/pycram/ros/ros2/service.py
```diff
@@ -0,0 +1,38 @@
...
+
+services = {}
+
+class ServiceProxy:
+    def __init__(self, topic_name, service_message):
+        self.service = node.create_client(service_message, topic_name)
+        self.message_type = service_message
+    def __call__(self, *args, **kwargs):
+        print(type(self.message_type.Request))
+        print(type(args[0]))
```
tomsch420, line 15: remove prints

src/pycram/ros_utils/object_state_updater.py
```diff
@@ -94,11 +92,11 @@ def __init__(self, tf_topic: str, joint_state_topic: str):
         :param joint_state_topic: Name of the joint state topic, needs to publish sensor_msgs/JointState
         """
         self.tf_listener = tf.TransformListener()
-        rospy.sleep(1)
+        time.sleep(1)
```
tomsch420, line 95: is time really correct here?

tomsch420 (changes requested):
So regarding the ros stuff, i think it would be nice to have an implementable interface instead  of modules containing functions. However, there is no real point since i dont believe in a ros3 like change soon where we would need another implementation.

The pinoccio stuff definitly needs a class.

src/pycram/external_interfaces/pinocchio_ik.py
Tigul: The Problem is that not just the error changes but also other Parts of the algorithm, for example the calculation of the jacobian. So I'm Not sure of just seperating the error makes sense

Tigul:
I don't think an Interface for the ROS stuff makes Sense I would Stick to the current approach with Modulen. Also because it makes it easier to expand other functionallity

src/pycram/external_interfaces/pinocchio_ik.py
```diff
@@ -48,10 +49,10 @@ def compute_ik(target_link: str, target_pose: Pose, robot: Object) -> Dict[str,
     # Initial joint configuration
     # q = pinocchio.neutral(model)
     q = create_joint_configuration(robot, model)
-    eps = 1e-4
-    IT_MAX = 10000
-    DT = 1e-1
-    damp = 1e-12
+    eps = PinocchioConfig.error_threshold
+    IT_MAX = PinocchioConfig.max_iterations
```
tomsch420, line 53: why not just take in an instance of pinochhio conf?

src/pycram/external_interfaces/pinocchio_ik.py
tomsch420: i see

tomsch420 (approved):

src/pycram/external_interfaces/pinocchio_ik.py
Tigul: Because then you need to always need to pass a config every time you call it

PR #272, opened 2025-02-16
Merge generic description

AbdelrhmanBassiouny:
Added merge description to generic object description, currently there are two methods:

One is by creating a URDF description from each and then merging them using the URDF description merge method.

The second is by merging their bounding box into a mesh and then creating a URDF description from it.

Also, fixed bounding box extents, and the duplicate application of scale when mesh_scale is specified in the Object constructor.


src/pycram/object_descriptors/generic.py
tomsch420, line 1: this is gucci. Perhaps there should be some subpackage of pycram called scene_preprocessing where we do these kind of things before runtime?

tomsch420 (approved):

Tigul (approved):

PR #274, opened 2025-02-19
Partial actions

Tigul:
# Description 
Action designator now inherit from PartialDesignators which allows to mix variables and list when parameterizing ActionDesignator. 
As an additional benefit will the PartialDesignator also evaluate  iterables when creating all possible combinations, meaning Designator as input will further evaluated when iterating all combination of input parameter of the ActionDesignator. 

## Exampe
The old syntax:
~~~
PickUpAction(milk_desig, [Arms.RIGHT], [Grasp.FRONT]).res...
~~~

With PartialDesignators
~~~
PickUpAction(milk_desig, [Arms.RIGHT, Arms.LEFT], Grasp.FONT).res...
~~~

### DIFF
--- a/src/pycram/datastructures/world.py
+++ b/src/pycram/datastructures/world.py
@@ -357,7 +357,7 @@ class World(WorldEntity, ABC):
         matching_objects = list(filter(lambda obj: obj.name == name, self.objects))
         return matching_objects[0] if len(matching_objects) > 0 else None
 
-    def get_object_by_type(self, obj_type: Type[PhysicalObject]) -> List[Object]:
+    def get_object_by_type(self, obj_type: PhysicalObject) -> List[Object]:
         """
         Return a list of all Objects which have the type 'obj_type'.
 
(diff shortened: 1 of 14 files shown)


src/pycram/datastructures/world.py
```diff
@@ -357,7 +357,7 @@ def get_object_by_name(self, name: str) -> Optional[Object]:
         matching_objects = list(filter(lambda obj: obj.name == name, self.objects))
         return matching_objects[0] if len(matching_objects) > 0 else None
 
-    def get_object_by_type(self, obj_type: Type[PhysicalObject]) -> List[Object]:
+    def get_object_by_type(self, obj_type: PhysicalObject) -> List[Object]:
```
tomsch420, line 360: are you sure about this? This means that one filters for an instance of PhysicalObject instead of the Type physical object and hence the name is missleading

Tigul:
Closing this for now since there will be a more refined PR soon

PR #275, opened 2025-02-20
Containment & Rules

AbdelrhmanBassiouny:
Added containment relations and an update method that updated the containment of given bodies.

This is based on the containment of bounding boxes.

Also added a SWRL rule in Owlready2 to assert that parents contain the objects contained by the parts.

Created a Rule class that inherits from HasConcept to facilitate the creation of rules.

Added a collection of rules for the world.

Added a couple of tests in test_links to test this functionality.


src/pycrap/ontologies/crax/rules.py
```diff
@@ -0,0 +1,80 @@
...
+
+class HierarchicalContainment(CRAXRule):
+    """
+    A rule that asserts that parent objects contain the objects contained by their parts.
+    """
+
+    @property
+    def rule_body(self):
+        return f"""
+        {is_part_of}(?part, ?parent), {contains_object}(?part, ?object)
```
tomsch420, line 78: heyyy very cool way of doing it!

tomsch420 (approved):

src/pycrap/ontologies/crax/rules.py
AbdelrhmanBassiouny: Thanks :smile: it took multiple iterations to reach this way of doing it haha.

PR #278, opened 2025-02-25
Multi-World Support for Pycram

mkhoshnam:
**world.py:**

- Modified logic so that _World.current_world_ is set only when _set_as_main_ is true and the world type is "**belief**"
- This prevents prospection worlds from inadvertently overriding the main belief world.
- Added new parameters _set_as_main_ and _world_type_ to the constructor.
- These parameters allow explicit designation of a world as a "**belief**" (main) world or a "**prospection**" (secondary or more) world.
- modified the __init_prospection_world_ and __sync_prospection_world_ methods to correctly handle the creation and synchronization of prospection worlds.

**world_factory.py:**

- **_WorldFactory_** class to create and manage multiple world instances.
- Each world is assigned a unique ID and registered in an internal registry for easy management.
- The factory distinguishes between "**belief**" and "**prospection**" worlds.
- Belief worlds are created with _set_as_main=True_, while prospection worlds are created with _set_as_main=False_.
-

### DIFF
--- a/src/pycram/datastructures/world.py
+++ b/src/pycram/datastructures/world.py
@@ -81,18 +81,19 @@ class World(WorldEntity, ABC):
     """
 
     def __init__(self, mode: WorldMode = WorldMode.DIRECT, is_prospection: bool = False, clear_cache: bool = False,
-                 id_: int = -1):
+                 id_: int = -1, set_as_main: bool = False, world_type: str = "belief"):
         """
-        Create a new simulation, the mode decides if the simulation should be a rendered window or just run in the
-        background. There can only be one rendered simulation.
-        The World object also initializes the Events for attachment, detachment and for manipulating the world.
+        Create a new simulation.
 
-        :param mode: Can either be "GUI" for rendered window or "DIRECT" for non-rendered. The default parameter is
-         "GUI"
-        :param is_prospection: For internal usage, decides if this World should be used as a prospection world.
-        :param clear_cache: Whether to clear the cache directory.
+        :param mode: Simulation mode (GUI/DIRECT).
+        :param is_prospection: True if this is a prospection world.
+        :param clear_cache: Whether to clear the cache.
         :param id_: The unique id of the world.
+        :param set_as_main: If True, set this world as the main current world.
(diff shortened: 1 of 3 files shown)


tomsch420:
Please do changes in the world instead of copying the entire file to a new one. Otherwise I cannot review what changed.

src/multipleworld/world_factory.py
```diff
@@ -0,0 +1,105 @@
...
+    worlds for easier management and lookup.
+    """
+
+    # Registry to hold created worlds keyed by their unique id.
+    _registry: Dict[int, World] = {}
+    # Internal counter for assigning unique ids.
+    _id_counter: int = 0
+
+    @classmethod
+    def create_world(cls, world_type: str, mode: WorldMode = WorldMode.DIRECT, clear_cache: bool = False,
```
AbdelrhmanBassiouny, line 21: It would be cleaner if world_type is a new Enum that has Prospection and Belief as members.

src/multipleworld/world.py
AbdelrhmanBassiouny, line 1: Why do we have two world.py?

AbdelrhmanBassiouny (commented):
Looks promising, but needs testing, if you could add tests for creating many worlds and test how they do not interfere with each other and all have different states, that would be great.

src/multipleworld/world_factory.py
```diff
@@ -0,0 +1,105 @@
+from typing import Dict, List, Optional
+from pycram.datastructures.enums import WorldMode
+from pycram.world import World
+
+
+class WorldFactory:
```
Tigul, line 6: I think the WorlFactory should be a singelton, since we only want exactly one. 
See here: https://python-patterns.guide/gang-of-four/singleton/ or here: https://github.com/cram2/pycram/blob/dev/src/pycram/local_transformer.py#L148 for implementation details

Tigul (commented):
I see some problems with the new handling of current_world, all reasoning methods use the current_world in combination with use_prospection_world to decide in which simulation the reasoning should be done. 
With the new way all reasoning will be done in the beliefe state

PR #279, opened 2025-03-06
Update return type to ORMAction

artnie:


artnie:
@Tigul the return types of these functions are complaining that its ORMAction, not Action.

Tigul:
Looks good

PR #280, opened 2025-03-09
Render task tree

AbdelrhmanBassiouny:
Added ```__str__()``` and ```__repr__()``` methods to all performables
Added a method ```render(filename: str)``` to the ```TaskTree```

### DIFF
--- a/src/pycram/designators/action_designator.py
+++ b/src/pycram/designators/action_designator.py
@@ -149,6 +149,22 @@ class ActionAbstract(ActionDesignatorDescription.Action, abc.ABC):
 
         return action
 
+    def __str__(self):
+        # all fields that are not ORM classes
+        fields = {}
+        for key, value in vars(self).items():
+            if key.startswith("orm_"):
+                continue
+            if isinstance(value, ObjectDesignatorDescription.Object):
+                fields[key] = value.name
+            elif isinstance(value, Pose):
+                fields[key] = value.__str__()
+        fields_str = "\n".join([f"{key}: {value}" for key, value in fields.items()])
+        return f"{self.__class__.__name__.replace('Performable', '')}:\n{fields_str}"
+
+    def __repr__(self):
+        return self.__str__()
+
 
 @dataclass
 class MoveTorsoActionPerformable(ActionAbstract):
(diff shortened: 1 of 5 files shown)


src/pycram/designators/action_designator.py
```diff
@@ -285,6 +291,12 @@ def validate(self, result: Optional[Any] = None, max_wait_time: timedelta = time
         if not validator.goal_achieved:
             raise ConfigurationNotReached(validator, configuration_type=StaticJointState.Park)
 
+    def __str__(self):
```
Tigul, line 294: You could make this more generic and move it to the super class, then it does not need to be added to every action individually

src/pycram/designators/motion_designator.py
```diff
@@ -95,6 +95,16 @@ def insert(self, session: Session, *args, **kwargs) -> ORMMoveTCPMotion:
 
         return motion
 
+    def __str__(self):
```
Tigul, line 98: Same here, move this to the super class

src/pycram/tasktree.py
```diff
@@ -243,6 +245,32 @@ def add_node(self, action: Optional[Action] = None) -> TaskTreeNode:
         self.current_node = new_node
         return new_node
 
+    @staticmethod
```
Tigul, line 248: Maybe add an option to print it to the terminal using the AsciiRenderer

tomsch420 (commented):
its fine. But why the overloaded repr and str methods? Isnt dataclasses doing that for you?

src/pycram/designators/action_designator.py
AbdelrhmanBassiouny: Yes would be better I guess, assuming we don't really need the action specific parameters in there (which I guess not)

AbdelrhmanBassiouny:
> its fine. But why the overloaded repr and str methods? Isnt dataclasses doing that for you?

the problem is the ORM attributes appear, and also I want to show just ObjectDescription.name not the whole thing, plus I want to print the parameters in different lines not all in one line as it makes nodes the tree so wide and unreadable.

src/pycram/designators/action_designator.py
AbdelrhmanBassiouny: Done

src/pycram/designators/motion_designator.py
AbdelrhmanBassiouny: Done

src/pycram/tasktree.py
AbdelrhmanBassiouny: Done

PR #282, opened 2025-03-09
Tasktree callback

AbdelrhmanBassiouny:
depends on #281 
Added the possibility to hook callbacks to task tree for specific action types.


tomsch420 (commented):
This raises a question why the task tree even has its own attributes anymore and is not just an aggregation object for performables. Lets discuss that tomorrow

AbdelrhmanBassiouny:
> There are already callbacks for pre and post perform of actions, how do these callbacks differ from these? Because they should be called at the same time during executio.

Good point, I haven't thought of those, but the thing is I found task tree doing exactly what I want like recording start and end timestamps, and building the tree knowing the action is part of which parent action...

PR #284, opened 2025-03-12
Readd gripper descriptions for TIAGo

J-Schaefer:
Add GripperType and opening_distance for the grippers.


Tigul:
looks good to me

PR #287, opened 2025-03-18
New actions

Tigul:
# New and revamped Designator 
This PR introduces a revamp of Action, Object and Location Designator. This now allows a more flexible way of handling input parameters and also to use Designators as inputs for other Designators, for example, a LocationDesignator as input for an ActionDesignator

## Changes in this PR 
* Action, Location and Object designator are combined with Partial Designator 
* ObjectDesignator return WorldObjects instead of the wrapper class
* LocationDesignator return Poses instead of the wrapper class 
* Action Designator do not have two classes anymore
    * There is only one class which has a ```description``` method that returns a PartialDesignator
* MoveTorsoAction now takes a TorsoState instead of joint_states
* Typing for all Designator classes 
* Renaming of Action Designators 

## Renaming of Actions 
Actions now only have one class which has a ```description``` method that returns a PartialDesignator, the ```description``` method is aliased for syntactic sugar. This replaces the old model of two classes the ```ActionDescription``` and the ```ActionPerformable```. 
What was the ```...ActionPerformable``` is now the ```...Action```. 
What was the ```...Action``` is now the ```...ActionDescription``` 

## Example 
Before: 
```
loc = CostmapLocation(....).resolve()
NavigateAction([loc.pose]).resolve().perform()
```

Now:
```
NavigateActionDescription(CostmapLocation(...)).resolve().perform()
```


tomsch420 (changes requested):
Up to a few things very well done!

src/pycram/designators/location_designator.py
```diff
@@ -105,59 +51,46 @@ class CostmapLocation(LocationDesignatorDescription):
...
-        List of arms with which the pose can be reached, is only used when the 'reachable_for' parameter is used
-        """
-
-        grasp_description: Optional[GraspDescription]
-        """
-        The grasp configuration used to reach the pose
+    def __init__(self, target: Union[Pose, Object],
+                 reachable_for: Optional[Union[Iterable[Object], Object]] = None,
+                 visible_for: Optional[Union[Iterable[Object], Object]] = None,
+                 reachable_arm: Optional[Union[Iterable[Arms], Arms]] = None,
```
LucaKro, line 57: okay here it looks like the variable should be "reachable_arms" instead

src/pycram/designators/motion_designator.py
```diff
@@ -314,8 +316,8 @@ def to_sql(self) -> ORMOpeningMotion:
 
     def insert(self, session: Session, *args, **kwargs) -> ORMOpeningMotion:
         motion = super().insert(session)
-        op = self.object_part.insert(session)
-        motion.object = op
+        # op = self.object_part.insert(session)
+        # motion.object = op
```
LucaKro, line 320: remove commented out code, if we need it back we have git

src/pycram/designators/motion_designator.py
```diff
@@ -346,8 +348,8 @@ def to_sql(self) -> ORMClosingMotion:
 
     def insert(self, session: Session, *args, **kwargs) -> ORMClosingMotion:
         motion = super().insert(session)
-        op = self.object_part.insert(session)
-        motion.object = op
+        # op = self.object_part.insert(session)
```
LucaKro, line 351: same as above

LucaKro (changes requested):

src/pycram/designators/action_designator.py
```diff
@@ -302,14 +316,29 @@ def validate(self, result: Optional[Any] = None, max_wait_time: timedelta = time
         if not validator.goal_achieved:
             raise ConfigurationNotReached(validator, configuration_type=StaticJointState.Park)
 
+    @classmethod
+    def description(cls, arm: Union[Iterable[Arms], Arms]) -> PartialDesignator[Type[ParkArmsAction]]:
+        return PartialDesignator(ParkArmsAction, arm=arm)
+
+
+SPECIAL_KNOWLEDGE = {
```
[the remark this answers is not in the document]
Tigul: Will be removed in a future PR by @LucaKro

src/pycram/designators/action_designator.py
```diff
@@ -730,48 +805,66 @@ class TransportActionPerformable(ActionAbstract):
     @with_tree
     def plan(self) -> None:
         robot_desig_resolved = BelieveObject(names=[RobotDescription.current_robot_description.name]).resolve()
-        ParkArmsActionPerformable(Arms.BOTH).perform()
+        ParkArmsAction(Arms.BOTH).perform()
         pickup_loc = CostmapLocation(target=self.object_designator,
                                      reachable_for=robot_desig_resolved,
-                                     reachable_arms=[self.arm],
+                                     reachable_arm=[self.arm],
```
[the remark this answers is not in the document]
Tigul: It takes an arm or an Iterble over arms but since action designator also have the singular as their signature I would like to keep this consistent

src/pycram/designators/location_designator.py
```diff
@@ -253,23 +253,23 @@ class AccessingLocation(LocationDesignatorDescription):
     Location designator which describes poses used for opening drawers
     """
 
-    def __init__(self, handle_desig: Union[ObjectDescription.Link, Iterable[ObjectDescription.Link]],
+    def __init__(self, handle: Union[ObjectDescription.Link, Iterable[ObjectDescription.Link]],
```
tomsch420, line 256: make the handle just a link?

Tigul: It is just a link though

PR #290, opened 2025-03-28
[Feature] Correct translation, generalized movehead, cleaned up PM

LucaKro:
# Feature: Generalized MoveHead ProcessModules and correct object-relative translations

Up until now, the move head calculations were faulty, using an incorrect formula as well as not accounting for joint offsets between the pan and tilt joint. This was fixed for the default process modules, which all robots except boxy and donbot can now use. Donbot now has its own process module for looking at a specific pose that utilises IK instead of four predefined static joint states. This now ensures that the robot always looks towards the correct pose, albeit with a fixed offset due to the offset between the joints and the camera frame, which may be addressed in the future.

In addition, transporting and related actions now consider the object type when grasping (a bowl, for example, has to be grasped at the rim, not in the middle of the object). Furthermore, an approach axis was introduced, which is determined by the front grasp (but can be overridden by setting it explicitly inside the robot description). The approach axis denotes the vector along which the robot should approach the object, relative to the object's rotation and the grasp description used.

## Highlights of This Pull Request

- Generalized MoveHead ProcessModules
- Generalized approaching of objects
- Object-specific grasp poses

---

### MoveHead ProcessModules

For robots utilizing the default movehead process module, the yaw and pitch joint of the neck now need to be set explicitly in the robot description. The orientation of the tilt joint is adjusted according to its <origin> tag inside the URDF, which ensures that we correctly calculate the desired joint state. The previous incorrect tilt formula was corrected, and now accurately calculates the angle of elevation:
`np.arctan2(adjusted_pose_in_tilt[2], np.sqrt(adjusted_pose_in_tilt[0] ** 2 + adjusted_pose_in_tilt[1] ** 2))`

Donbot, on the other hand, calculates a pose above its shoulder joint (for good reachability with its camera, mounted on top of the gripper) that is oriented towards the target. Then, an IK solution to move the camera to the desired pose is calculated and applied

### Approaching of objects

We now approach objects along an approach axis, which is automatically derived from the first row of the front grasp rotation matrix (unless it was explicitly set in the robot description). This vector is then turned into a transform applied to a pose. Here, it is expected that the pose's quaternion was primed for grasping previously by applying the grasp quaternion to the pose.

### Object-specific grasp poses
Unique grasp poses (in this case, a rim grasp) can be approached to objects in the ontology. The method "get_grasp_pose" for a world entity then queries the information and implements what should happen with the entity pose. In the case of the bowl, we take the width of the bowl mesh, divide it by two, and use the result to guess where the rim of the bowl is. 
The pre-pose distance is also calculated dynamically based on the largest dimension of the object mesh.

---

## Backwards-Incompatible Changes

- Any prepose-distance parameter needs to be removed from costmaps and pick/place actions
- Robots using the default Move-Head process modules must specify the yaw/pitch joints in the robot description. I have already edited the existing ones, but if you are using your own description/robot, you may need to adjust this

---

Credit: Layout inspired by [SciPy](https://docs.scipy.org/doc/scipy/release/1.15.0-notes.html)


src/pycram/world_concepts/world_object.py
```diff
@@ -1340,23 +1340,30 @@ def get_joint_parent_link(self, joint_name: str) -> ObjectDescription.Link:
         """
         return self.joints[joint_name].parent_link
 
-    def find_joint_above_link(self, link_name: str) -> str:
+    def find_joint_above_link(self, link_name: str, joint_type: JointType = None) -> str | None:
```
tomsch420, line 1343: str | None is not supported by typing below 3.10. Use optional. Also typing of joint_type is wrong.

tomsch420 (changes requested):

tomsch420 (approved):

tomsch420 (approved):

src/pycram/datastructures/world_entity.py
```diff
@@ -581,3 +581,35 @@ def calculate_grasp_descriptions(self, robot: Object) -> List[GraspDescription]:
         grasp_configs = objectTmap.calculate_grasp_descriptions(robot, preferred_grasp_alignment)
 
         return grasp_configs
+
+    def get_grasp_pose(self, end_effector, grasp: GraspDescription) -> Pose:
```
Tigul, line 585: If this does not affect the orientation then maybe rename it to get_grasp_position

src/pycram/designators/location_designator.py
```diff
@@ -222,7 +217,7 @@ def __iter__(self):
 
         final_map = self.setup_costmaps(target, test_robot)
 
-        with UseProspectionWorld():
+        with (UseProspectionWorld()):
```
Tigul, line 220: Why?

src/pycram/local_transformer.py
```diff
@@ -231,6 +231,24 @@ def get_object_from_frame(self, frame: str) -> Optional[Object]:
         found_objects = [obj for obj in world.objects if frame == obj.tf_frame]
         return found_objects[0] if len(found_objects) > 0 else self.get_object_from_link_frame(frame)
 
+    def translate_pose_along_local_axis(self, pose: Pose, axis: List, distance: float) -> Pose:
```
Tigul, line 234: We have an enum for axis.

This function has not really to do with the local transformer so I'm Not sure if this is the right place.

src/pycram/local_transformer.py
```diff
@@ -231,6 +231,24 @@ def get_object_from_frame(self, frame: str) -> Optional[Object]:
...
+        :param pose: The pose that should be translated
+        :param axis: The local axis along which the translation should be performed
+        :param distance: The distance by which the pose should be translated
+
+        :return: The translated pose
+        """
+        transform = pose.to_transform("self")
+        normalized_translation_vector = np.array(axis) / np.linalg.norm(axis)
+        scaled_translation_vector = (normalized_translation_vector * distance).tolist()
+        translation = Transform(scaled_translation_vector, [0, 0, 0, 1], "self", "translation",
```
Tigul, line 248: You can also just multiply the Translations without the Transform conversion and use the Frame from the Pose otherwise the returned pose will be in the wrong Frame.

src/pycram/process_modules/pr2_process_modules.py
```diff
@@ -1,3 +1,5 @@
+import actionlib
```
Tigul, line 1: Please remove this Import since it will fail in ROS2, use the Abstraction in pycram.ros instead

src/pycram/robot_description.py
```diff
@@ -268,12 +274,14 @@ def get_manipulator_chains(self) -> List[KinematicChainDescription]:
...
 
-    def get_camera_frame(self, robot_object_name: str) -> str:
+    def get_camera_frame(self, robot_object_name: str = None) -> str:
         """
         Quick method to get the name of a link of a camera. Uses the first camera in the list of cameras.
 
         :return: A name of the link of a camera
         """
+        if robot_object_name is None:
+            return self.name
```
Tigul, line 284: Return also the Camera Frame Here otherwise the function propose changes

src/pycram/robot_description.py
```diff
@@ -312,6 +320,19 @@ def get_static_joint_chain(self, kinematic_chain_name: str, configuration_name:
             raise ValueError(f"There is no KinematicChain with name {kinematic_chain_name} for robot {self.name}. "
                              f"The following chains are available: {list(self.kinematic_chains.keys())}")
 
+    def get_offset(self, name):
+        """
+        Returns the offset of a Joint in the URDF.
```
Tigul, line 325: Newline

src/pycram/robot_description.py
```diff
@@ -312,6 +320,19 @@ def get_static_joint_chain(self, kinematic_chain_name: str, configuration_name:
             raise ValueError(f"There is no KinematicChain with name {kinematic_chain_name} for robot {self.name}. "
                              f"The following chains are available: {list(self.kinematic_chains.keys())}")
 
+    def get_offset(self, name):
```
Tigul, line 323: Typing

src/pycram/robot_description.py
```diff
@@ -312,6 +320,19 @@ def get_static_joint_chain(self, kinematic_chain_name: str, configuration_name:
...
+        Returns the offset of a Joint in the URDF.
+        :param name: The name of the Joint for which the offset will be returned.
+        :return: The offset of the Joint
+        """
+        if name not in self.urdf_object.joint_map.keys():
+            logerr(f"The name: {name} is not part of this robot URDF")
+            return None
+
+        offset = self.urdf_object.joint_map[name].origin
+        return offset if offset else None
```
Tigul, line 334: [0, 0, 0], [0, 0, 0, 1] would be better than None

src/pycram/robot_description.py
```diff
@@ -385,6 +406,34 @@ def get_arm_chain(self, arm: Arms) -> Union[KinematicChainDescription, List[Kine
...
+        joints along with their corresponding links.
+
+        :param yaw_joint: The joint name for the yaw movement of the neck.
+        :param pitch_joint: The joint name for the pitch movement of the neck.
+        :param roll_joint: The joint name for the roll movement of the neck.
+        """
+        yaw_link = self.get_child(yaw_joint) if yaw_joint else None
+        pitch_link = self.get_child(pitch_joint) if pitch_joint else None
+        roll_link = self.get_child(roll_joint) if roll_joint else None
+        self.neck = {
```
Tigul, line 422: Why Not Filter the None values

Tigul (changes requested):

src/pycram/datastructures/world_entity.py
```diff
@@ -581,3 +581,35 @@ def calculate_grasp_descriptions(self, robot: Object) -> List[GraspDescription]:
...
+
+        if self.ontology_concept.has_preferred_alignment[0].has_rim_grasp[0].value:
+            approach_axis = end_effector.get_approach_axis()
+            approach_direction = grasp.approach_direction
+            rim_direction = GraspDescription(approach_direction, None, False)
+            rim_direction_index = approach_direction.value[0].value.index(1)
+            rim_offset = self.get_rotated_bounding_box().dimensions[rim_direction_index] / 2
+            grasp_pose.rotate_by_quaternion(end_effector.grasps[rim_direction])
+            grasp_pose = LocalTransformer().translate_pose_along_local_axis(grasp_pose, approach_axis, -rim_offset)
+            grasp_pose = Pose(grasp_pose.position_as_list(), self.orientation_as_list)
```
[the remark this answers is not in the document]
LucaKro: the last line? yes it does, it resets the orientation back to the original one, since it was rotated in line 604. we need 604 to get the direction we need to know the rim direction, but its not the actual grasp orientation. that's why I reset the orientation back before returning the pose, so that its trivial to apply the grasp orientation afterwards

src/pycram/datastructures/world_entity.py
LucaKro: im not sure, when saying "get_grasp_position" i would expect a return of vector3 describing the new position.

but what actually happens is that we adjust the object pose, to get the grasp pose. I think just because the orientation doesn't change, doesn't mean that the name doesn't fit

src/pycram/designators/location_designator.py
LucaKro: uh tbh not sure, must have been inserted automatically when I clicked "refactored file" maybe? i didn't do this by hand. do you want me to remove it? it shouldn't make any difference

src/pycram/local_transformer.py
LucaKro: where would you recommend this method to sit? I originally had this inside the pose class, but after talking to tom I came to the conclusion that it fits here better.

also what enum do we have for this? you mean the axis identifier? axis identifier only describes default values for x,y or z. the axis we give may be any 3dvector

src/pycram/robot_description.py
LucaKro: you mean only include it in the dictionary if its not none?

src/pycram/datastructures/world_entity.py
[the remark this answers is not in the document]
Tigul: Oh yeah I overlookd the self. my bad

src/pycram/local_transformer.py
Tigul: Yeah this is better than pose, I would say either utils or we let it sit her although I'm not very happy with that. 

Yes the axisIdentifier, it gives you a vector describing the axis you could also adjust the enum to represent any axis.

src/pycram/robot_description.py
Tigul: Yes because the return is a little weired otherwise and user have to filer for that when using the function.

test/test_designator/test_action_designator.py
```diff
@@ -181,6 +181,7 @@ def test_transport(self):
         with simulated_robot:
             action_designator.MoveTorsoActionDescription([TorsoState.HIGH]).resolve().perform()
             description.resolve().perform()
+            description.resolve().perform()
```
Tigul, line 184: Two times?

Tigul (approved):

PR #298, opened 2025-05-08
[Perception] fixed detect and small changes on pouring

sunava:
wip i will rework this but for now its a quick fix


Tigul (approved):
There are two tests failing, one because Detect doesn't define the frozen object and the other because the  skipped validation. 
Can you skip the tests for now?

PR #316, opened 2025-05-28
Lazy monitors

Tigul:
# Description 
Some small improvements: 
    * Added a condition in case the ik algorithm gets unstable 
    * Improved monitors to have more behavior, they can now Interrupt, pause and resume if the condition is true
    * Improved Error messages from lazy_product

### DIFF
--- a/src/pycram/language.py
+++ b/src/pycram/language.py
@@ -6,7 +6,7 @@ from queue import Queue
 from typing_extensions import Iterable, Optional, Callable, Dict, Any, List, Union, Tuple, Self, Sequence, Type, \
     TYPE_CHECKING
 
-from .datastructures.enums import TaskStatus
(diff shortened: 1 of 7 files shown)


src/pycram/language.py
```diff
@@ -380,43 +386,53 @@ class MonitorNode(SequentialNode):
...
-        monitor_thread.join()
+        self.monitor_thread.join()
 
     def monitor(self):
         while not self.kill_event.is_set():
             if self.condition.get_value():
-                self.interrupt()
+                if self.behavior == MonitorBehavior.INTERRUPT:
+                    self.interrupt()
+                elif self.behavior == MonitorBehavior.PAUSE:
```
tomsch420, line 432: this smells like something where dependency injection would solve these switch cases. Is it a better solution here to just pass the callable in as behavior?

PR #325, opened 2025-06-03
added Carry Action and Vector3dStamped

sunava:


src/pycram/process_modules/default_process_modules.py
```diff
@@ -432,11 +433,11 @@ def _execute(self, designator: MoveTCPMotion):
         lt = LocalTransformer()
         pose_in_map = lt.transform_pose(designator.target, "map")
         tip_link = RobotDescription.current_robot_description.get_arm_chain(designator.arm).get_tool_frame()
-        root_link = "map"
+        root_link = "base_link"
```
Tigul, line 436: I don't think that is a good idea since this are the default_process_modules

src/pycram/process_modules/pr2_process_modules.py
```diff
@@ -77,8 +77,14 @@ def feedback_callback(msg):
             pass
 
         goal = Pr2GripperCommandGoal()
-        goal.command.position = 0.0 if designator.motion == GripperState.CLOSE else 0.1
-        goal.command.max_effort = 50.0
+
+        position_map = {
```
Tigul, line 81: That should be in the robot_description

sunava: well this wasnt in the description before i can move it later there

PR #333, opened 2025-06-16
Fix Dockerfile + requirements.txt

rhaschke:
- Fix warnings
- Install xacro, which is missing as a dependency in pulled iai sources
  Better should be fixed there, of course.
- Use local sources

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,14 +1,12 @@
 gitpython>=3.1.37
 pycram_bullet==3.2.8
-pathlib~=1.0.1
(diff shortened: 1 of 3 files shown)


requirements.txt
Tigul, line 1: Why did you remove numpy? And did you test this? because the differentiation of the python version is necessary for Ubuntu24.04

rhaschke: I think, there is no need to fix the numpy version. It is pulled in via other dependency, e.g. pinocchio.
Only tested on Ubuntu 22.04.

PR #337, opened 2025-06-26
updated EfficientTransportAction

joy1304:
This action selects the closest available arm using Euclidean distance.

### DIFF
--- a/.github/workflows/update-docker-image.yml
+++ b/.github/workflows/update-docker-image.yml
@@ -5,7 +5,7 @@ on:
       - master
       - dev
     paths:
-      - 'requirements.txt'
+      - 'docker/ros2/Dockerfile'
 # only run when a commit has changes in the requirements.txt file
 
 jobs:
--- a/docker/ros2/Dockerfile
+++ b/docker/ros2/Dockerfile
@@ -11,6 +11,7 @@ SHELL ["/bin/bash", "-c"]
 
 RUN apt update && apt install python3.12-venv ros-jazzy-xacro python3-vcstool git ros-dev-tools default-jre -y
 RUN vcs import --input https://raw.githubusercontent.com/cram2/pycram/dev/rosinstall/pycram-ros2-https.rosinstall
+RUN git clone -b ros-jazzy https://github.com/code-iai/iai_robots.git
 RUN source /opt/ros/jazzy/setup.bash && cd $OVERLAY_WS && colcon build --symlink-install
 RUN echo "source $OVERLAY_WS/install/setup.bash" >> ~/.bashrc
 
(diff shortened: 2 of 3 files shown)


src/pycram/designators/action_designator.py
```diff
@@ -783,6 +783,134 @@ def description(cls, object_designator: Union[Iterable[Object], Object],
         return PartialDesignator(TransportAction, object_designator=object_designator,
                                  target_location=target_location,
                                  arm=arm)
+    
+
+from dataclasses import dataclass
+from pycram.designators.action_designator import ActionDescription, with_plan
+from pycram.designators.object_designator import ObjectDesignatorDescription
+from pycram.designators.location_designator import CostmapLocation
+from pycram.designators.action_designator import (
```
Tigul, line 792: This import causes a circular import since you are from the file in itself. Also move the imports to the top of the file and merge with the existing imports

src/pycram/designators/action_designator.py
```diff
@@ -783,6 +783,134 @@ def description(cls, object_designator: Union[Iterable[Object], Object],
...
+        robot = BelieveObject(names=[RobotDescription.current_robot_description.name]).resolve()
+        obj = self.object_designator.resolve()
+
+        if not obj or not obj.pose:
+            raise ConfigurationNotReached(f"Couldn't resolve the pose for the object: {self.object_designator}")
+
+        object_position = obj.pose.position  # (x, y, z)
+
+        # The gripper positions for arms
+        left_tip = robot.get_link_position("left_gripper")
```
Tigul, line 826: This only works for tracy since the names are different for other robots. Please use the RobotDescription to get the tool_frames

src/pycram/designators/action_designator.py
```diff
@@ -783,6 +783,134 @@ def description(cls, object_designator: Union[Iterable[Object], Object],
...
+        right_tip = robot.get_link_position("right_gripper")
+
+        # Euclidean distances from each gripper to the object
+        def euclidean(p, q):
```
Tigul, line 830: Why define the euclidian distance yourself and not use np.linalg.norm?

src/pycram/designators/action_designator.py
```diff
@@ -783,6 +783,134 @@ def description(cls, object_designator: Union[Iterable[Object], Object],
...
+        ParkArmsActionDescription(Arms.BOTH).perform()
+
+    @classmethod
+    @with_plan
+    def description(cls, object_designator, target_location):
+        """
+        This creates a designator for EfficientTransportAction, which automatically
+        picks the best free arm based on proximity — no need to specify the arm manually.
+        """
+        return PartialDesignator(
```
Tigul, line 907: Please look at the other actions on how to initialize the PartialDesignator

Tigul (changes requested):
This code is similar to #332 maybe it is better to combine both and integrate the arm choosing into the pick and place action

PR #342, opened 2025-07-03
Refactor imports to use the 'rx' module in probabilistic_model

benjaminalt:
A rename in probabilistic_model had caused import errors on a fresh install of PyCRAM.


Tigul:
Please increase the minimum required version of probabilistic models

PR #343, opened 2025-07-03
Add missing scikit-learn dependency to requirements.txt

benjaminalt:
Unittests won't run without it


Tigul:
SK-Learn is a dependency of probabilistic_model and not PyCRAM

PR #350, opened 2025-08-07
First addition of multi-robot plans

pkehr:
## Summary

This PR introduces the `RobotManager` class, which enables managing and controlling multiple robots within a world environment.

## Changes

- Replaces direct usage of `World.robot` and `RobotDescription.current_robot_description` with:
  - `RobotManager.get_active_robot()`
  - `RobotManager.get_robot_description()`
- Enables specifying which robot to use inside a `with_robot` statement by passing the respective robot object.  
  **Example:** `with_real_robot(pr2_robot)`
- Adds a test case that swaps and verifies the active robot.


src/pycram/multirobot/robot_manager.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def __init__(self):
+        """
+        Initialize RobotManager
+        """
+        rdm = RobotDescriptionManager()
+        logdebug(f'Initialize RobotManager.')
+
+    @staticmethod
+    def add_robot(robot):
```
LucaKro, line 53: typing

src/pycram/multirobot/robot_manager.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        Add another robot to the list of available robots
+
+        :param robot: Robot object to add
+        """
+        RobotManager.available_robots[robot.name] = robot
+
+    @staticmethod
+    def set_active_robot(robot_name: str = None):
```
LucaKro, line 62: maybe input robot and then get its name? its confusing imo when some method take robot, some take its name as a string etc

LucaKro (changes requested):
some small clarifications and method signature typing, but otherwise good work

src/pycram/multirobot/robot_manager.py
```diff
@@ -0,0 +1,124 @@
...
+    @staticmethod
+    def get_active_robot(robot=None):
+        """
+        Returns the active robot if the given variable is None.
+        This differentiation is used for handling multiple robots in multithreaded Actions
+
+        :param robot: Robot object
+        :return: Currently active robot
+        """
+        robot = robot or RobotManager.active_robot
```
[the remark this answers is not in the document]
pkehr: This change is part of concurrent multi-robot execution, where multiple robots can be controlled simultaneously. For that system it is not sufficient to only have one _active_robot_ and is therefore passed as an argument / returned directly.
In systems where this does not matter (single-robot and sequential multi-robot) the parameter "robot" shall not be passed at all, resulting in using RobotManager.active_robot.

src/pycram/multirobot/robot_manager.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        Retrieve the description of the active robot.
+        For concurrent robot actions, a robot can be given as parameter and its description is used instead.
+
+        :param robot: Robot object to use. If None is given, use the first robot added instead.
+
+        return: RobotDescription
+        """
+        if robot is None:
+            return RobotDescription.current_robot_description
```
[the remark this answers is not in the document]
pkehr: This is similar to the RobotManager.active_robot discussion.
In single-robot or sequential multi-robot scenarios, this variable is unused, since only the description of the active robot is required.
But in concurrent situations where multiple robots act simultaneously, this approach is insufficient and direct robot information is required.
The remaining implementation for concurrent execution will be included in a future PR.

src/pycram/multirobot/robot_manager.py
```diff
@@ -0,0 +1,131 @@
...
+from pycram.robot_description import RobotDescriptionManager, RobotDescription
+from pycram.ros import logdebug
+
+if TYPE_CHECKING:
+    from pycram.world_concepts.world_object import Object
+
+
+class RobotManager(ABC):
+    """
+    Class for managing multiple robots simultaneously
```
LucaKro, line 16: Mention here that its a singleton

src/pycram/process_module.py
```diff
@@ -98,7 +98,9 @@ def __exit__(self, _type, value, traceback):
         ProcessModuleManager.execution_type = self.pre
         ProcessModule.execution_delay = self.pre_delay
 
-    def __call__(self):
+    def __call__(self, robot=None):
```
LucaKro, line 101: typing, the other \_\_call\_\_ methods as well

LucaKro (changes requested):

LucaKro (approved):

